PR #19, opened 2022-02-23
Implement TRPR0008 and fix the Guilty value

bjpirt:


src/types/Plea.ts
```diff
@@ -0,0 +1,8 @@
+export enum Plea {
+  Guilty = 1,
+  NotGuilty = 2,
+  Consents = 4,
+  Guilty6 = 6,
```
MihaiMadeT, line 5: is guilty6 used anywhere?

MihaiMadeT (approved):
one comment

daviesjamie (approved):

src/types/Plea.ts
bjpirt: Good spot - this came from the data in Bichard and I added it for completeness. Basically, Guilty can be either `1` or `6`

PR #39, opened 2022-03-21
Import comparison from e2e tests and add command to test them all

bjpirt:


emadkaramad (approved):

tomvaughan77 (approved):

tests/comparison.test.ts
```diff
@@ -0,0 +1,37 @@
...
+      const exceptions = extractExceptionsFromAho(annotatedHearingOutcome)
+
+      it("should match triggers", () => {
+        expect(coreResult.triggers).toStrictEqual(triggers)
+      })
+
+      it("should match exceptions", () => {
+        expect(coreResult.exceptions).toStrictEqual(exceptions)
+      })
+    } catch (e) {
```
jaskaransarkaria, line 31: Should we have a test for error cases?

jaskaransarkaria (approved):

tests/comparison.test.ts
bjpirt: What do you mean? At the moment if we throw an error this will show up as a failing test

PR #42, opened 2022-03-22
Update comparison tests to fix some of the differences

bjpirt:


tests/comparison.test.ts
```diff
@@ -1,22 +1,27 @@
...
   .readdirSync(filePath)
   .map((name) => `${filePath}/${name}`)
   .map(processTestFile)
   .filter((t) => !t.incomingMessage.match(/<br7:HearingOutcome/))
 
+const filter = process.env.FILTER_TEST
+if (filter) {
+  tests = tests.filter((t) => t.file.includes(filter))
+}
+
```
jaskaransarkaria, line 20: Can tests be a const and then include this append this filter to line 15? This feels a little odd.

tests/helpers/generateMockPncQueryResultFromAho.ts
```diff
@@ -0,0 +1,154 @@
...
+        "@_PNCID": string
+      }
+      CourtCases: {
+        CourtCase: AhoPncCourtCase | AhoPncCourtCase[]
+      }
+    }
+  }
+}
+
+const parsePncDate = (pncDate: string): Date | null => {
```
jaskaransarkaria, line 56: Could we parse the date with zod here?

jaskaransarkaria (approved):

tests/comparison.test.ts
jaskaransarkaria: `.filter((t) => process.env.FILTER_TEST ? t.file.includes(filter) : true)`

PR #50, opened 2022-03-25
Add some validations for the defendant structure

bjpirt:
Need to wait for the PNC enrichment before we can test some of these validations


bjpirt:
Figured it's best to get this merged in before waiting for PNC amendments so it doesn't block things

emadkaramad (approved):

PR #53, opened 2022-03-29
feat: 💡 port the offence structures over to typescript and rejig the

jaskaransarkaria:
* Adds the Offence structure amend (we call it "enrich") code from Bichard.
* rejigs the types to use discriminated unions

Note errors will be thrown when we validate later.


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "The code to replace the processing logic of Bichard 7",
   "main": "build/index.js",
   "scripts": {
-    "test": "jest --maxWorkers=2 --testPathIgnorePatterns='comparison'",
+    "test": "jest --maxWorkers=2 --testPathIgnorePatterns='comparison' --testNamePattern",
```
jaskaransarkaria, line 7: allows you to run a single test quickly `npm run test <regex for test>`

src/lib/parseOffenceReason.test.ts
```diff
@@ -0,0 +1,90 @@
...
+          __type: "CommonLawOffenceCode",
+          CommonLawOffence: COMMON_LAWS,
+          Reason: "ML001",
+          Qualifier: "A",
+          FullCode: `${COMMON_LAWS}001A`
+        }
+      })
+    })
+
+    it.skip("THEN parse an offenceCode with a INDICTMENT prefix", () => {
```
jaskaransarkaria, line 21: Skipped because there are no Indictment codes in the `offence-code.json` so the test will fail as it can't be confirmed against the json.

src/lib/parseASN.ts
```diff
@@ -0,0 +1,22 @@
+import type { ParsedASN } from "src/types/ParsedASN"
```
bjpirt, line 1: FYI: there's already some code for parsing an ASN - see `lib/asn.ts`. Though to be fair, that code doesn't handle 21 char ASNs

bjpirt (approved):

bjpirt:
Please could you create a ticket to add the exceptions that were skipped in this PR?

PR #55, opened 2022-03-29
Bicaws 1947 enrich offence results post pnc enrichment

emadkaramad:
This PR is for enriching offences results after PNC enrichment:
- Populates court
- Populates source organisation
- Populates PNC disposal
- Populates result class


src/exceptions/HO100322.ts
```diff
@@ -0,0 +1,9 @@
+import type { ExceptionGenerator } from "src/types/ExceptionGenerator"
+
+const HO100322: ExceptionGenerator = (hearingOutcome) => {
```
bjpirt, line 3: Happy to do these exceptions in a second pass - please could you add tickets to make sure we don't forget?

bjpirt (approved):

src/exceptions/HO100322.ts
emadkaramad: I was actually going to do the exceptions in my second PR as part of the same ticket. The logic is relevant to the amendment and should be easier to implement now.

PR #66, opened 2022-04-06
BICAWS-1951: Validate offence structure

tomvaughan77:
There is 2 parts to this PR:

1. Add tests around legacy bichard to capture the exceptions it throws around offences, the approach here was to test incoming xml with invalid data and capture exceptions it threw. (see HO100232/ OffenceTitle & HO100234/ OffenceWording)

2. With new bichard we drift from legacy bichard's behaviour by validating and throwing exceptions where legacy bichard couldn't, we used the `validator.xml` in legacy bichard as our guide.


src/lib/isCaseRecordable.ts
```diff
@@ -4,7 +4,7 @@ const isCaseRecordable = (annotatedHearingOutcome: AnnotatedHearingOutcome): boo
   const isPncQuery = annotatedHearingOutcome.PncQuery !== undefined
   const isThereARecordableOffence =
     annotatedHearingOutcome.AnnotatedHearingOutcome.HearingOutcome.Case.HearingDefendant.Offence.some(
-      (offence) => offence.RecordableOnPNCindicator
+      (offence) => offence.RecordableOnPNCindicator === "Y"
```
jaskaransarkaria, line 7: This was previously set to a boolean in zod however in legacy bichard `validator.xml` this is an Enum that comes from data `[yes-no.json](https://github.com/ministryofjustice/bichard7-next-data/blob/main/output-data/data/yes-no.json)`

emadkaramad (approved):

PR #67, opened 2022-04-12
test: 💍 add `it.ifNewBichard` to conditionally run tests

jaskaransarkaria:
Some exceptions are hidden by legacy bichard; with new bichard we have the ability to raise these exceptions during validation.

In such cases we don't/ can't test these exceptions in legacy bichard but do want to test for these exceptions in new bichard (conditional testing) using the new syntax:

`it.ifNewBichard()`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,7 @@
 module.exports = {
   root: true,
   env: { es6: true },
-  ignorePatterns: ["build/*"],
+  ignorePatterns: ["build/*", "jest.setup.ts"],
   overrides: [
     {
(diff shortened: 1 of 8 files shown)


tomvaughan77:
This looks fab 🙌 Any chance you can also replace the `it.skips()` tests that are already in the repo with this?

jaskaransarkaria:
> This looks fab raised_hands Any chance you can also replace the `it.skips()` tests that are already in the repo with this?

I did try but they are actually failing at the minute think we should have a separate ticket to clean those ones up

PR #72, opened 2022-04-29
BICAWS-1875 Use native types 

C-gyorfi:
### Changes:
- Update @moj-bichard7-developers/bichard7-next-data package
- Use boolean types for `RecordableOnPNCindicator` and `NotifiableToHOindicator` in AHO schema. These fields indicating boolean values but previously used strings to represent true or false
- Update snapshots


src/use-cases/enrichHearingOutcome/enrichFunctions/enrichOffences/enrichOffence.ts
```diff
@@ -15,7 +15,7 @@ const enrichOffence = (offence: Offence): Offence => {
 
   if (offenceIgnored) {
     offence.OffenceCategory = "B7"
-    offence.RecordableOnPNCindicator = "N" // other possible options are "D" (don't know), "NA" (not applicable)
+    offence.RecordableOnPNCindicator = false // other possible options are "D" (don't know), "NA" (not applicable)
```
tomvaughan77, line 18: If `RecordableOnPNCindicator` has other possible options, does it belong as an enum instead?

tests/exceptions/HO100236.test.ts
```diff
@@ -24,10 +24,10 @@ describe("HO100233", () => {
       cjsCode: "MC8080524",
       offenceCategory: "CB",
       offenceTitle: "Application to reopen case",
-      recordableOnPnc: "N",
+      recordableOnPnc: false,
       description: "blah",
       homeOfficeClassification: "123/45",
-      notifiableToHo: "Y",
+      notifiableToHo: false,
```
tomvaughan77, line 30: Should be `true`

C-gyorfi: Ooops, that is not looking good

C-gyorfi: I'll have a look

C-gyorfi: Good catch, updated, although I found it was irrelevant for the test

src/use-cases/enrichHearingOutcome/enrichFunctions/enrichOffences/enrichOffence.ts
C-gyorfi: Looks like we only use it as a boolean(Y or N) in the core repo, not sure about old Bichard 😕

C-gyorfi: Removed comment in [d043f17](https://github.com/ministryofjustice/bichard7-next-core/pull/72/commits/d043f17a672fa22ed44b4cdd4840e28ceac0bb49)

tomvaughan77 (approved):

tomvaughan77 (approved):

PR #75, opened 2022-05-24
BICAWS-1905: Generate AHO XML

tomvaughan77:
First-pass converting internal `AHO` to XML and exporting it when new Bichard is run.

Mapping is still not 100% accurate due to a combination of missed fields/enrichments, bugs, and mappings that haven't appeared in the ~3 test files we've run this on so far. It's a start though, and we can hopefully use these xml comparisons to tell us what we have/haven't completed in the core rewrite.


src/index.ts
```diff
@@ -7,16 +9,20 @@ import parseSpiResult from "./use-cases/parseSpiResult"
 import transformSpiToAho from "./use-cases/transformSpiToAho"
 
 export default (message: string, pncGateway: PncGateway): BichardResultType => {
+  let hearingOutcome: AnnotatedHearingOutcome
```
monotypical, line 12: nit: we can declare this `const` on line 15, it's value doesn't (and shouldn't) change after assigning

src/lib/generateAhoXml.ts
```diff
@@ -0,0 +1,362 @@
...
+    "ds:DurationLength": d.DurationLength
+  }))
+
+const mapAhoResultsToXml = (results: Result[]): Br7Result[] =>
+  results.map((result) => ({
+    "ds:CJSresultCode": result.CJSresultCode,
+    "ds:SourceOrganisation": mapAhoOrgUnitToXml(result.SourceOrganisation),
+    "ds:CourtType": result.CourtType,
+    "ds:ResultHearingType": { "#text": result.ResultHearingType, "@_Literal": "Other" },
+    "ds:ResultHearingDate": result.ResultHearingDate ? format(result.ResultHearingDate, "yyyy-MM-dd") : "",
```
monotypical, line 65: suggestion: should we leave the XML fields out when things don't exist on the AHO, instead of having fields present with the empty string value? It will make the code handing this XML more complex when it has to check for fields that exist but aren't empty, and imo it would be more of a "true" representation to have these fields be optional in the types and not have the fields present in the XML when they're not in the AHO, but there might be some context on this I'm missing

monotypical (approved):

C-gyorfi (approved):

C-gyorfi:
Maybe we could tidy up the commit history or just squash before merging

src/index.ts
tomvaughan77: Totally agree. However this is an artefact of the `parsing` side of this ticket that didn't get cleaned up when they split. Going to leave it for now, since it's just about to be changed to this anyway

src/lib/generateAhoXml.ts
tomvaughan77: Good point, have done this now. I left along any keys that map to attributes in the xml (anything that looks like `@_thing`) as these seem to want to be empty strings.

monotypical (approved):

monotypical (approved):

PR #76, opened 2022-05-25
Jamie/bicaws 1946 pnc validate and amend

bjpirt:
It's a big one - we've ported across all of the logic from Bichard for identifying which court case each offence belongs to. The code is pretty gnarly and once we have added the tests (separate ticket once we've handled XML parsing since the tests use that) it will be due a _massive_ refactor.

What this code does:
- The main aim of this code is to assign the court case reference number that exists in the PNC to the offences incoming from the court
- For some reason this is not already in the data, and we need it because that's how we know what to update in the PNC later in the process
- It tries to match by sequence number, then by offence code, start date, end date and result codes. Gradually getting less precise until it finds a match.
- It also looks across multiple court cases in the PNC to try and find the match

We've tried to keep the processing logic the same as original Bichard to make it easier to make sure we've actually got all of the logic. We've also ported the tests where they exist so that we can be sure that our implementation does actually match. There are a few things still outstanding that can go into subsequent tickets:
- Handling penalty cases in the PNC (we don't do this yet)
- Adding the paths to the exceptions properly (at the moment they're just placeholders)
- Porting across the top level tests (they use AHO XML files which we are currently in the process of handling properly)


daviesjamie (approved):
https://c.tenor.com/aAsftu-ZFLUAAAAC/ron-swanson-parks-and-rec.gif

PR #85, opened 2022-06-09
Ho100306

jaskaransarkaria:


src/use-cases/dataLookup/lookupOffenceByCjsCode.ts
```diff
@@ -0,0 +1,58 @@
+import { offenceCode } from "@moj-bichard7-developers/bichard7-next-data"
+
+import type { OffenceCode } from "@moj-bichard7-developers/bichard7-next-data/dist/types/types"
+import { ExceptionCode } from "src/types/ExceptionCode"
+import type { LookupResult } from "src/types/LookupResult"
+
+export const lookupOffenceByCjsCode = (cjsCode: string): OffenceCode | undefined =>
+  offenceCode.find((x) => x.cjsCode === cjsCode)
+
+export const lookupNationalOffenceByCjsCode = (cjsCode: string, areaCode?: string): LookupResult => {
```
bjpirt, line 10: I'm not sure whether we should be returning exception objects from here as they are tightly coupled to the domain model. We coul dreturn different types of errors and check for them where we call this. Open to discussion though because it's tricky due to the fact that different exceptions can be raised for different types of failure

jaskaransarkaria: > We coul dreturn different types of errors and check for them where we call this. Open to discussion though because it's tricky due to the fact that different exceptions can be raised for different types of failure

I don't think it's tied to domain other than the object key is called exception (which I can happily change to error if preferred?). It's a pattern used quite commonly in fp, wrapping the error in a Monad. I think it is a purer and more functional way of handling exceptions.

Benefits:
  - avoid blowing up the function throwing an error and then try/catching
  - easily testable
  - purer and more predictable function

tests/exceptions/HO100108.test.ts
```diff
@@ -46,7 +46,7 @@ describe("HO100108", () => {
       expectTriggers: false
     })
 
-    // Check the right triggers are generated
+    // Check the right exceptions are generated
```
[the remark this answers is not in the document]
jaskaransarkaria: That's a good point @C-gyorfi, I agree with you.

bjpirt (approved):

PR #105, opened 2022-06-21
Implement HO100305 and fix some bugs that were causing it to be incorrectly generated

bjpirt:
Also took the opportunity to convert Result Class to an enum


jaskaransarkaria (dismissed):

tomvaughan77 (approved):
Looks great! Do we also want to add a characterisation test for this though? (ignore if not)

C-gyorfi (dismissed):

emadkaramad (approved):

PR #133, opened 2022-07-06
Improve cli comparison tool

bjpirt:
```
Options

  -f, --file string     Specify either the local file path or an S3 URL                               
  -s, --start string    Specify the start timestamp in ISO8601 format                                 
  -e, --end string      Specify the end timestamp in ISO8601 format                                   
  -p, --filter string   Filter based on the last result. Specify either 'failure', 'success', 'both'. 
                        Default is 'failure'                                                          
  -h, --help            Prints this usage guide   
```
Screenshot 2022-07-06 at 11 14 22

### DIFF
--- a/src/comparison/cli.ts
+++ b/src/comparison/cli.ts
@@ -1,3 +1,3 @@
 import main from "./cli/main"
 
-main().then(() => {})
+main()
--- a/src/comparison/cli.ts
+++ b/src/comparison/cli.ts
@@ -1,36 +1,3 @@
-import fs from "fs"
-import compare from "src/comparison/compare"
-import { formatXmlDiff } from "src/comparison/xmlOutputComparison"
+import main from "./cli/main"
 
-const FILE_PATH = process.argv.slice(-1)[0]
-
-const contents = fs.readFileSync(FILE_PATH).toString()
-const result = compare(contents, true)
-
-if (result.debugOutput) {
-  if (!result.triggersMatch) {
-    console.log("Triggers do not match")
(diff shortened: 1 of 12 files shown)


src/comparison/cli.ts
```diff
@@ -1,36 +1,3 @@
...
-  if (!result.xmlParsingMatches) {
-    console.log("XML parsing does not match")
-    console.log(formatXmlDiff(result.debugOutput.xmlParsingDiff))
-  }
-}
-
-if (result.triggersMatch && result.exceptionsMatch && result.xmlOutputMatches && result.xmlParsingMatches) {
-  console.log("Comparison matches")
-}
+main().then(() => {})
```
daviesjamie, line 3: Do you need the `.then` chained on here?

daviesjamie (commented):
Could we add a script to the `package.json`? So instead of 

```
npx ts-node src/comparison/compare.ts
```

We could run something like

```
npm run compare
```

src/comparison/cli.ts
bjpirt: I thought we did because it's an async function but it runs fine without so I'll remove

bjpirt:
> Could we add a script to the `package.json`? So instead of
> 
> ```
> npx ts-node src/comparison/compare.ts
> ```
> 
> We could run something like
> 
> ```
> npm run compare
> ```

There's already a `compare` command but I think this one should probably replace it so I'll update

daviesjamie (dismissed):

daviesjamie (approved):

PR #144, opened 2022-07-18
Add GitHub Action to publish npm package

daviesjamie:
This PR adds a **manually triggered** GitHub Actions workflow that will build and publish the npm package.

Specifically, when manually triggered on the main branch, it will:

1. Check to make sure the version number in `package.json` hasn't already been published. If it has, it will fail the workflow with a helpful message saying to bump the version number.
2. Publish the package to npm using the [shared bichard7 user](https://www.npmjs.com/~bichard7)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,16 @@
 
 The code to replace the processing logic of Bichard 7
 
+## Publishing package updates
+
+The code in this repository is packaged in the [`@moj-bichard7-developers/bichard7-next-core` NPM package](https://www.npmjs.com/package/@moj-bichard7-developers/bichard7-next-core).
+
+To deploy a new version of the package:
+
+1. Manually bump the version number in `package.json` in your PR. It's recommended to follow [semantic versioning principles](https://semver.org).
+1. Merge your PR into the `main` branch.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -2,6 +2,16 @@
...
 The code to replace the processing logic of Bichard 7
 
+## Publishing package updates
+
+The code in this repository is packaged in the [`@moj-bichard7-developers/bichard7-next-core` NPM package](https://www.npmjs.com/package/@moj-bichard7-developers/bichard7-next-core).
+
+To deploy a new version of the package:
+
+1. Manually bump the version number in `package.json` in your PR. It's recommended to follow [semantic versioning principles](https://semver.org).
+1. Merge your PR into the `main` branch.
```
tomvaughan77, line 12: nit: Should be `2.` and `3.`

tomvaughan77 (approved):

README.md
daviesjamie: Disagree 😁 

Markdown will render multiple `1.` as the correct number in the sequence (check the [preview](https://github.com/ministryofjustice/bichard7-next-core/blob/ef0c84c684fa41f84a415ca8637f3e382dfeae63/README.md)), but using `1.` makes it easier to rearrange/insert points in the future.

https://google.github.io/styleguide/docguide/style.html#use-lazy-numbering-for-long-lists

tomvaughan77: Lush, didn't know it did that 😂 Ty

PR #188, opened 2022-08-25
Bicaws 2237 add audit logger

emadkaramad:
This PR adds and injects audit logger to core logic.

For more info about the functions of audit logger, check the `AuditLogger` interface.


tomvaughan77 (approved):

src/types/AuditLogEvent.ts
```diff
@@ -0,0 +1,4 @@
+export default interface AuditLogEvent {
```
bjpirt, line 1: We might need to make the definition of an event match the ones we store in the Audit Log API but we could do this at a later date

emadkaramad: Agreed. I left it for the next PR when we are adding the existing audit logs to Core.

PR #190, opened 2022-08-26
Import and cache standing data modules when needed

emadkaramad:
In this PR:
The` requireStandingData() `function is updated to import and cache standing data modules when needed.
This reduces memory usage when invoking comparison lambda. (We only need one of the standing data versions when invoking comparison lambda)

Webpack config is also updated to compile standing data modules in separate files. Codebuild artificats script is updated to include the standing data files in the zip file.


jaskaransarkaria (approved):
Amazing 🎉

jaskaransarkaria (approved):

PR #224, opened 2022-11-08
feat: 🎸 convert aho to hearing out xml with no validations

jaskaransarkaria:
For resubmit functionality in the new ui we need to generate from an aho some xml without validations (eg. hasErrors="false")

* pass validate parameter (defaults to true) to `convertAhoToXml()` and sub functions
* conditionally apply the `hasErrors` property based on the validate parameter


src/serialise/ahoXml/generate.ts
```diff
@@ -262,16 +263,18 @@ const mapAhoResultsToXml = (
...
-      "AnnotatedHearingOutcome",
-      "HearingOutcome",
-      "Case",
-      "HearingDefendant",
-      "Offence",
-      offenceIndex,
-      "Result",
-      resultIndex
-    ]),
+    ...(validate && {
```
bjpirt, line 266: might be tidier to pass validate into hasError and just return undefined?

src/serialise/ahoXml/generate.ts
```diff
@@ -563,8 +584,10 @@ const convertAhoToXml = (hearingOutcome: AnnotatedHearingOutcome): string => {
   }
 
   const builder = new XMLBuilder(options)
-  const xmlAho = mapAhoToXml(hearingOutcome)
-  addExceptionsToAhoXml(xmlAho, hearingOutcome.Exceptions)
+  const xmlAho = mapAhoToXml(hearingOutcome, validate)
+  if (validate) {
+    addExceptionsToAhoXml(xmlAho, hearingOutcome.Exceptions)
```
bjpirt, line 589: I wonder about moving all of the `hasError` calls into `addExceptionsToAhoXml`

src/serialise/ahoXml/generate.ts
```diff
@@ -216,12 +187,8 @@ const mapNextResultSourceOrganisation = (
...
 
-const mapAhoResultsToXml = (
-  results: Result[],
-  exceptions: Exception[] | undefined,
-  offenceIndex: number
-): Br7Result[] =>
-  results.map((result, resultIndex) => ({
+const mapAhoResultsToXml = (results: Result[]): Br7Result[] =>
+  results.map((result) => ({
     "ds:CJSresultCode": text(result.CJSresultCode.toString()),
```
jaskaransarkaria, line 192: @bjpirt Is the incorrect casing intentional here? It's the same in the aho schema too

src/serialise/ahoXml/generate.ts
```diff
@@ -533,26 +479,39 @@ const mapAhoCXE01ToXml = (pncQuery: PncQueryResult): Cxe01 => ({
     : undefined
 })
 
-const mapAhoToXml = (aho: AnnotatedHearingOutcome): AhoXml => {
+const mapAhoToXml = (aho: AnnotatedHearingOutcome, validate = true): AhoXml => {
```
bjpirt, line 482: I don't think the typing is correct for this function now - if validate is false, it will return a `Br7AnnotatedHearingOutcome` I think

src/serialise/ahoXml/generate.ts
```diff
@@ -533,26 +479,39 @@ const mapAhoCXE01ToXml = (pncQuery: PncQueryResult): Cxe01 => ({
...
+            CXE01: aho.PncQuery ? mapAhoCXE01ToXml(aho.PncQuery) : undefined,
+            "br7:PNCQueryDate": aho.PncQueryDate ? toISODate(aho.PncQueryDate) : undefined,
+            "br7:PNCErrorMessage": optionalText(aho.PncErrorMessage),
+            ...xmlnsTags
+          }
+        }
+      : {
+          ...hearingOutcome
+        })
   } as AhoXml
```
bjpirt, line 511: I think we should remove this manual casting so the compiler can be sure we're returning the right thing

src/serialise/ahoXml/generate.ts
jaskaransarkaria: It does seem to fit the pre-existing type:

```
export interface AhoXml {
  "?xml": XML
  "br7:AnnotatedHearingOutcome"?: Br7AnnotatedHearingOutcome
  "br7:HearingOutcome"?: Br7HearingOutcome
}
```

bjpirt (approved):

PR #325, opened 2023-03-17
Add basic typescript api

ian-antking:


api/src/index.ts
```diff
@@ -0,0 +1,5 @@
+import app from "./app"
+
+const PORT: string = process.env.PORT || "4000"
```
C-gyorfi, line 3: I wonder if we need to pick a different port, the e2e test docker container runs on port 4000

api/src/routes/status.ts
```diff
@@ -0,0 +1,8 @@
+import { Router } from "express"
+import { checkStatus } from "../controllers/status"
+
+const router: Router = Router()
+
+router.get("/", checkStatus)
```
C-gyorfi, line 6: Can we add a test for this endpoint?

C-gyorfi (approved):
Approving with a few comments

api/src/routes/status.ts
ian-antking: added a basic integration test with supertest

C-gyorfi (approved):

api/src/app.ts
```diff
@@ -0,0 +1,10 @@
+import * as express from "express"
+import * as cors from "cors"
+import statusRouter from "./routes/status"
+
+const app: express.Application = express()
+app.use(cors())
+
+app.use("/status", statusRouter)
```
emadkaramad, line 8: If we rename it to `/health` we can then use it as a health endpoint for our ecs services.

emadkaramad (approved):

emadkaramad (dismissed):

.eslintrc.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   root: true,
   env: { es6: true },
-  ignorePatterns: ["build/*", "jest.setup.ts"],
+  ignorePatterns: ["build/*", "api/build/*", "jest.setup.ts"],
```
bjpirt, line 4: Wondering whether we should just completely ignore everything in the api folder and give it its own config for ts, eslint, etc?

bjpirt (changes requested):

.eslintrc.js
C-gyorfi: I think it would be good to have one linting for the whole repo, assuming we'd use the same setting in both projects

emadkaramad (dismissed):

bjpirt (dismissed):

C-gyorfi (approved):

PR #334, opened 2023-03-23
Add buildspec and dockerise API

C-gyorfi:


bjpirt:
Approved, but we should consider moving to esbuild in line with the other projects in the future

bjpirt (approved):

emadkaramad (approved):

PR #370, opened 2023-05-05
Bicaws-2730: Add api endpoint for listing cases

ian-antking:
This P.R is to add an endpoint for listing court cases.

- Ported over typeORM from the UI
- Kept the interface the same as the UI
- We expect queries to be in a query string
- We switched to `tsc` due to `esbuild` not working as intended - no real impact on build time at this stage


api/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+```bash
+curl http://localhost:3333/court-cases?forces%5B0%5D=01&maxPageItems=10&allocatedToUserName=username&caseState=Resolved&courtDateRange%5B0%5D%5Bfrom%5D=2023-05-02T15%3A06%3A58.412Z&courtDateRange%5B0%5D%5Bto%5D=2023-05-02T15%3A06%3A58.412Z&courtName=courtName&defendantName=defendantName&locked=false&order=asc&orderBy=defendantName&pageNum=1&ptiurn=ptirun&reasonCode=reason&reasons%5B0%5D=Bails&resolvedByUsername=username&urgent=Urgent
+```
+
+Example response:
+
+```json
+{
+  "result": [],
```
C-gyorfi, line 57: Should we include an example of the court case object?

api/src/services/entities/transformers/resolveFindOperator.ts
```diff
@@ -0,0 +1,21 @@
+import { FindOperator } from "typeorm"
+
+const resolveFindOperator = <TInput, TOutput>(
```
C-gyorfi, line 3: What does this transformer do? Do we need to unit test it to make sure it behaves as expected?

api/src/services/listCourtCases.ts
```diff
@@ -0,0 +1,214 @@
...
+  const pageNumValidated = (pageNum ? parseInt(pageNum, 10) : 1) - 1 // -1 because the db index starts at 0
+  const maxPageItemsValidated = maxPageItems ? parseInt(maxPageItems, 10) : 25
+  const repository = connection.getRepository(CourtCase)
+  const subquery = connection
+    .getRepository(Note)
+    .createQueryBuilder("notes")
+    .select("COUNT(note_id)")
+    .where("error_id = courtCase.errorId")
+  let query = repository.createQueryBuilder("courtCase")
+  query = courtCasesByVisibleForcesQuery(query, forces) as SelectQueryBuilder<CourtCase>
```
C-gyorfi, line 42: FYI there were [some changes](https://github.com/ministryofjustice/bichard7-next-ui/pull/243/files) to `courtCasesByVisibleForcesQuery`, and also added filter for [excluded triggers.](https://github.com/ministryofjustice/bichard7-next-ui/pull/250/files). Maybe not in this, but It would be good to update this in a follow up PR!

api/tests/court-cases.integration.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  describe("GET", () => {
+    it("returns a 200 status code", async () => {
+      const response = await request(app)
+        .get("/court-cases")
+        .query(stringify({ forces: ["01"], maxPageItems: "10" }))
+
+      expect(response.statusCode).toBe(200)
+    })
+
+    it("returns a 400 status code if required query attributes are missing", async () => {
```
C-gyorfi, line 26: Could we add a test for 500 error?

api/src/controllers/courtCases.ts
```diff
@@ -0,0 +1,20 @@
+import type { Response } from "express"
+import type { DataSource } from "typeorm"
+import getDataSource from "../services/getDataSource"
+import listCourtCases from "../services/listCourtCases"
+import type { CaseListQueryRequest } from "../../src/types/CaseListQueryRequest"
+
+export const getCourtCases = async (req: CaseListQueryRequest, res: Response) => {
```
C-gyorfi, line 7: Nit: the file name does not match the function name

api/src/controllers/courtCases.ts
```diff
@@ -0,0 +1,20 @@
...
+  const { caseListQueryParams } = req
+  const dataSource: DataSource = await getDataSource()
+  try {
+    const data = await listCourtCases(dataSource, caseListQueryParams!)
```
C-gyorfi, line 11: What would happen if caseListQueryParams is undefined?

C-gyorfi (dismissed):
Nice work, approving with some comments!

api/src/controllers/courtCases.ts
ian-antking: listCourtCases would throw an error. The validator middleware would bounce it back before this stage though, so I thought it was safe enough.

api/src/controllers/courtCases.ts
ian-antking: My intention here was to have all of the controllers related to courtCases in this file.

api/src/services/entities/transformers/resolveFindOperator.ts
ian-antking: This is ported directly from the UI, so we really just treated it as a library.

api/README.md
ian-antking: 👍

C-gyorfi (approved):

PR #402, opened 2023-06-16
Add Architecture Diagrams

richrace:
- Added a static site generator and interactive one
- Uses docker for the tools
- Added npm commands
- Added README for how to run commands in the doc path

structurizr-1-HybridBichard (1)

structurizr-1-Container-001-key


environment/docker-compose-arch.yml
```diff
@@ -0,0 +1,22 @@
+version: "3.2"
```
bjpirt, line 1: Wonder if this should live in the docs folder as it's not part of the environment for running Core?

bjpirt (dismissed):

environment/docker-compose-arch.yml
richrace: 👍  Good spot. Leftover from initial spike.

PR #423, opened 2023-07-05
Refactor: Define const for audit log event options and use enum for event source

C-gyorfi:
### Context
This change was [suggested in a previous PR in the UI repo](https://github.com/ministryofjustice/bichard7-next-ui/pull/318#discussion_r1247905306). The UI repo uses the core repo package.

### Changes
- Define a const for all possible audit log event options(event type and code)
- Use a const for eventSource and eventCategory

### Next step
Use these types/const in the UI repo

### DIFF
--- a/src/types/AuditLogEvent.ts
+++ b/src/types/AuditLogEvent.ts
@@ -23,131 +23,131 @@ export type AuditLogEventOption = {
 }
 
 export const AuditLogEventOptions = {
-  HEARING_OUTCOME_RECEIVED_PHASE_1: {
+  hearingOutcomeReceivedPhase1: {
     code: "hearing-outcome.received-phase-1",
     type: "Hearing outcome received by phase 1"
   },
-  HEARING_OUTCOME_RECEIVED_PHASE_2: {
+  hearingOutcomeReceivedPhase2: {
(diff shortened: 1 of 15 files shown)


src/types/AuditLogEvent.ts
```diff
@@ -12,4 +12,143 @@ type AuditLogEvent = {
...
+  EnrichWithPncQuery = "EnrichWithPncQuery"
+}
+
+export type AuditLogEventOption = {
+  code: string
+  type: string
+}
+
+export const AuditLogEventOptions = {
+  HEARING_OUTCOME_RECEIVED_PHASE_1: {
```
emadkaramad, line 26: I think the json keys should be camelCase here to be consistent and follow the naming convention.

C-gyorfi: Yeah that was a lazy copy-paste from old repo, hoped I can get away with it 😄

C-gyorfi: Fixed in [430b13e](https://github.com/ministryofjustice/bichard7-next-core/pull/423/commits/430b13e0464f5641b7cc638c61cd8f1f5e636e7f)

emadkaramad (approved):

PR #434, opened 2023-07-11
Core npm package files

emadkaramad:
This PR is to
- transpile to cjs format using esbuild and resolve import paths
- generate types using `tsc`


build.package.js
```diff
@@ -0,0 +1,57 @@
...
+  await esbuild.build({
+    ...buildOptions,
+    entryPoints: files,
+    outdir: "dist",
+    minify: false,
+    bundle: false
+  })
+
+  console.log("Resolving import paths...")
+  await resolveImportPaths()
```
emadkaramad, line 49: I couldn't find a way to resolve alias paths when importing modules in esbuild. This function updates paths that start with `src/` with a relative path.

build.package.js
```diff
@@ -0,0 +1,57 @@
...
+      } else if (absolute.endsWith(`.${extension}`) && !absolute.endsWith(`.test.${extension}`)) {
+        files.push(absolute)
+      }
+    })
+  )
+
+  return files
+}
+
+async function resolveImportPaths() {
```
emadkaramad, line 21: Unfortunately, I couldn't find a straightforward solution for this. I've written a function that updates paths starting with `src/` by transforming them into relative paths.

richardcane (approved):

PR #489, opened 2023-08-31
AHO documentation in GitHub pages

emadkaramad:


.github/workflows/publish-architecture.yml
```diff
@@ -10,7 +10,7 @@ permissions:
 
 jobs:
   deploy_github_pages:
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
```
richrace, line 13: The reason why I was building it on `macos` was because I was getting Graphviz issues. 

We can enable the workflow by removing the `branches` below

```
on:
  push:
    branches:
      - main
```

richrace (approved):

PR #517, opened 2023-09-22
Add proposal for using DTO in UI

richrace:


docs/architecture/adrs/0002-use-dto.md
```diff
@@ -0,0 +1,36 @@
...
+  - `DisplayPartial` - Prefix for the object which has limited fields
+  - `DisplayFull` - Prefix for the object which has all fields (that aren't a security risk)
+- File structure for DTOs
+  - `src/services/dto` top level
+  - Then for example: `objectDto.ts`
+- Which holds functions like:
+  - `objToDisplayObjDto`
+  - `objToDisplayPartialObjDto`
+
+A DTO object is a new object mirror some of a Entity's attributes. We are actively using `Pick<T, "field">` to select
```
bjpirt, line 28: a new object mirror `with` some

docs/architecture/adrs/0002-use-dto.md
```diff
@@ -0,0 +1,36 @@
...
+what we want to use out of the Entity and mapping via a function (e.g. `const displayObj = objToDisplayObjDto(obj)`).
+
+We can control the vary levels of data we want share. E.g. `displayPartialObj` or `displayFullObj`.
```
bjpirt, line 31: "can control the vary levels" -> "can control the levels"

docs/architecture/adrs/0002-use-dto.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Consequences
+
+- It's more complex than just using the Database Record
```
bjpirt, line 35: - It's more secure than sending out raw database records into the UI

bjpirt (approved):

docs/architecture/adrs/0002-use-dto.md
richrace: Do you want to add it in or replace line 35

bjpirt: Add it in I think

emadkaramad (approved):

docs/architecture/adrs/0002-use-dto.md
```diff
@@ -0,0 +1,38 @@
...
+
+Using a Data Transfer Object (DTO) in the `bichard-next-ui` repo will help separate the User Interface (React) from
+the Database objects. It will also allow to use TypeScript more effectively, enforcing stricter types on objects at
+compile time.
+
+We have chosen the naming convention of:
+
+- Types:
+  - `Display` - Prefix for the object to be consumed in the UI
+  - `DisplayPartial` - Prefix for the object which has limited fields
```
monotypical, line 19: We might need to redact different fields in different contexts, for example when showing who left a note we only need the user's name, whereas when looking at who has a case locked we need to also know the user's groups.

Is the intention here to have a single `DisplayPartial` for e.g. a user? I think it would be worth mentioning how we would handle having multiple partial objects for each database object for different contexts, or clarifying the intention to have a single partial object.

richrace: > We might need to redact different fields in different contexts, for example when showing who left a note we only need the user's name, whereas when looking at who has a case locked we need to also know the user's groups

That has been done for the notes see: https://github.com/ministryofjustice/bichard7-next-ui/blob/main/src/services/dto/noteDto.ts#L17C28-L17C28

> Is the intention here to have a single DisplayPartial for e.g. a user? I think it would be worth mentioning how we would handle having multiple partial objects for each database object for different contexts, or clarifying the intention to have a single partial object.

You can have as many DTOs as you wish. E.g. `DisplayPartialUser`, `DisplayPartialUserForContext` would be a valid use case. Ideally, we want to remove as much data from the Objects so that they only have the bare minimum required to work.

monotypical (dismissed):

emadkaramad (approved):

C-gyorfi (dismissed):

richardcane (dismissed):

emadkaramad (approved):

PR #622, opened 2024-01-08
Check for MQ mode before looking at conductor

richardcane:


bjpirt (dismissed):

packages/message-forwarder/src/forwardMessage/forwardMessage.ts
```diff
@@ -21,30 +21,33 @@ const forwardMessage = async (message: string, client: Client): PromiseResult<vo
     throw new Error(`Unsupported destination type: "${destinationType}"`)
   }
 
-  const aho = parseAhoXml(message)
-  if (isError(aho)) {
-    throw aho
+  const maybeAHO = parseAhoXml(message)
```
richardcane, line 24: Just a rename

packages/message-forwarder/src/forwardMessage/forwardMessage.ts
```diff
@@ -21,30 +21,33 @@ const forwardMessage = async (message: string, client: Client): PromiseResult<vo
...
+  if (isError(maybeAHO)) {
+    throw maybeAHO
+  }
+  const correlationId = maybeAHO.AnnotatedHearingOutcome.HearingOutcome.Hearing.SourceReference.UniqueID
+
+  if (destinationType === DestinationType.MQ) {
```
richardcane, line 30: Move this up so that if we're in MQ mode, we don't bother checking Conductor (it doesn't exist in legacy mode in CI, which is what breaks everything)

bjpirt (approved):

PR #630, opened 2024-01-29
Increase note_text length limit to 2000 characters in the database

Senny10:
This PR is to increase the default note text length in the database. The migration was tested against the creation of 5 million records, which resulted in 33.20 seconds.
image
image
image
image


environment/postgres/liquibase/migrations/base/033_note_length_limit_increase.sql
```diff
@@ -0,0 +1,5 @@
+--liquibase formatted sql
+--changeset sen:033_note_length_limit_increase runInTransaction:false
```
bjpirt, line 2: This should be run in a transaction so remove `runInTransaction:false`

bjpirt (approved):

PR #654, opened 2024-02-28
Switch to using MailHog for smtp testing

bjpirt:
We previously used an in-process fake smtp server that was very unreliable. This PR switches this out for a docker container running [MailHog](https://github.com/mailhog/MailHog) which has an API you can check against for emails sent. One of the nice upsides of this is that you can visit http://localhost:8025 and get a webmail interface for things like the login codes for Bichard.

Screenshot 2024-02-28 at 16 22 15


packages/conductor/e2e-test/incomingMessageHandler.e2e.test.ts
```diff
@@ -5,27 +5,22 @@ process.env.S3_AWS_SECRET_ACCESS_KEY = "test"
...
 
 describe("Incoming message handler", () => {
-  let mailServer: MockMailServer
-
-  beforeAll(() => {
-    mailServer = new MockMailServer(20002)
-  })
-
-  afterAll(() => {
-    mailServer.stop()
```
joe-fol, line 28: Does MailHog have its own built in stop function?

Senny10 (approved):

joe-fol (approved):

packages/conductor/e2e-test/incomingMessageHandler.e2e.test.ts
bjpirt: Good spot - this should be removed

bjpirt: mailHog runs as a separate process and is integrated into the Docker environment now so we don't stop it as part of the tests

PR #692, opened 2024-04-02
Adds phase 2 processing to conductor compareFiles

ksugden:


packages/core/comparison/conductor-tasks/compareFiles.ts
```diff
@@ -37,24 +54,32 @@ const compareFiles: ConductorWorker = {
...
+      const phaseResults = nonErrorTestResults.filter((res) => res.phase === phase) as ComparisonResult[]
+      const gateway = new DynamoGateway(createDynamoDbConfig(phase))
+      const recordPhaseResults = await recordResultsInDynamo(phaseResults, gateway)
+      if (isError(recordPhaseResults)) {
+        return failed(`Failed to write phase ${phase} results to Dynamo`)
+      }
+    })
 
-    logs.push(`Results of processing: ${count.pass} passed. ${count.fail} failed`)
+    logs.push(
```
bjpirt, line 78: The majority of the time this will only handle a single message so it might be worth conditionally building the log message because most of the time it will print `undefined` for one of the phases

ksugden: I've added a function to build only relevant logs

Senny10 (approved):

PR #722, opened 2024-05-22
Phase 2 - Turn AHO to PNCUpdateDataset

emadkaramad:


packages/core/phase1/lib/errorPaths.ts
```diff
@@ -5,13 +5,23 @@ const offencePath = (offenceIndex: number): ExceptionPath => [
   "HearingOutcome",
   "Case",
   "HearingDefendant",
-  "Offence",
-  offenceIndex
+  ...(offenceIndex < 0 ? [] : ["Offence", offenceIndex])
```
emadkaramad, line 8: When offence index is -1, it means Result is not associated with and offence. (HearingDefendant > Result)

bjpirt (approved):

PR #729, opened 2024-05-23
add test coverage for isMatchToPncAdjAndDis

ian-antking:


ksugden (commented):
Just wondering if we need to handle the case of getCourtCaseFromQueryResults returning an array instead of a single case.  What if more than one case matched the courtCaseRef?

ksugden (approved):

PR #743, opened 2024-05-30
Enable the message forwarder to shut down gracefully

bjpirt:


chubberlisk (approved):
:shipit:

packages/message-forwarder/src/server.e2e.auto.test.ts
```diff
@@ -28,13 +28,15 @@ const resubmittedAho = fs.readFileSync("src/test/fixtures/success-exceptions-aho
 describe("Server in auto mode", () => {
   let messageData: string
   let correlationId: string
+  let mf: MessageForwarder
```
chubberlisk, line 31: A bit of nitpick, but I'd prefer to be explicit even though we're in the context of the message forwarder package, just to remove that smol cognitive load. It means `await messageForwarder.start()` reads nicer too.

```suggestion
  let messageForwarder: MessageForwarder
```

bjpirt: That's fair - will change now

richrace (approved):

PR #790, opened 2024-06-18
Bicaws7 2875 - Update generateTriggers to generate post-update triggers

Senny10:
In this PR:
- Added `phase` parameter to `generateTriggers` function. This parameter is used to decide wether post-update triggers should be generated.
- Also, updated `generateTriggers` to pass `options` parameter to each trigger function.
- Fixed an issue with `getRange` function in `DynamoGateway`. When there was a page with no result after filtering, it was breaking the loop and not retrieving the remaining pages.
- Removed `createTriggerIfNecessary` function
- Removed `identifyPreUpdateTriggers` function
- Removed `combineTriggerLists` function
- Removed `mostSpecificCourtRuleAllowsTrigger` function
- Removed `mostSpecificForceRuleAllowsTrigger` function
- Removed `identifyPostUpdateTriggers` function


packages/core/phase1/triggers/generate.ts
```diff
@@ -15,11 +17,22 @@ const independentTriggerFn = (aho: AnnotatedHearingOutcome) =>
...
 ): Trigger[] => {
-  const generatedTriggers = generator(aho, { triggers: existingTriggers })
-  const filteredTriggers = filterExcludedTriggers(aho, generatedTriggers)
+  const generatedPreUpdateTriggers = generator(aho, { triggers: existingTriggers })
+  const filteredPreUpdateTriggers = filterExcludedTriggers(aho, generatedPreUpdateTriggers)
+
+  let generatedPostUpdateTriggers: Trigger[] = []
+  let filteredPostUpdateTriggers: Trigger[] = []
+  if (generatePostUpdateTriggers && "PncOperations" in aho) {
+    generatedPostUpdateTriggers = identifyPostUpdateTriggers(aho)
```
emadkaramad, line 30: This is temporary until we move the post-update triggers logic to separate trigger files. Once it's done, generator function will generate both pre and post update triggers.

packages/core/phase1/triggers/generate.ts
```diff
@@ -30,11 +43,16 @@ const generateSetOfTriggers = (
   return existingTriggers.concat(finalTriggers)
 }
 
-export default (annotatedHearingOutcome: AnnotatedHearingOutcome): Trigger[] => {
+export default (annotatedHearingOutcome: AnnotatedHearingOutcome, generatePostUpdateTriggers = false): Trigger[] => {
```
emadkaramad, line 46: `generatePostUpdateTriggers` is temporary until we move the pos-update triggers code to separate trigger files. This is to ensure that we don't break Phase 1.

bjpirt: Why don't we just do this now so we avoid too many changes to phase 1?

bjpirt (dismissed):

packages/core/phase1/triggers/generate.ts
```diff
@@ -1,24 +1,26 @@
...
-import type { TriggerGenerator } from "../types/TriggerGenerator"
+import type { TriggerGenerator, TriggerGeneratorOptions } from "../types/TriggerGenerator"
 
-const independentTriggerFn = (aho: AnnotatedHearingOutcome) =>
+const independentTriggerFn = (aho: AnnotatedHearingOutcome, options?: TriggerGeneratorOptions) =>
   Object.entries(triggers)
     .filter(([triggerCode]) => triggerCode != TriggerCode.TRPR0015 && triggerCode != TriggerCode.TRPR0027)
     .reduce((acc: Trigger[], [_, trigger]) => {
-      return acc.concat(trigger(aho))
+      return acc.concat(trigger(aho, options))
```
emadkaramad, line 14: It seems `options` parameter was missing. Updated to pass the missing argument.

bjpirt (approved):

PR #795, opened 2024-06-21
Refactor parameters of intentional difference functions

chubberlisk:
## Context

We're looking to add a new intentional difference for Phase 2 comparison tests, but noticed that the current functions for identifying intentional differences have a number of parameters, some which are used in some, whilst in others only a few.

- Add types to use as the parameter for intentional differences.
  - This means a function can just destruct what they need to use instead of using `_` when an parameter isn't needed, which is noisy.
  - `BichardComparisonOutput` and `CoreComparisonOutput` types have been defined for readability purposes, so that it's clear what is `expected` and `actual`. Open to feedback about the naming of these (or anything as usual)!
- Remove unused `matchingToFinalOffences` intentional difference.
- Refactor all intentional difference functions to use new type.

## Next steps

1. Update intentional differences implementation to determine whether one is applicable for Phase 1 and/or Phase 2.
2. Update Phase 2 comparison tests to include checking of intentional differences.
3. Add intentional difference for when Bichard incorrectly raises a HO200114 exception, when the sequence numbers don't match exactly e.g. "2" and "002".


packages/core/comparison/lib/isIntentionalDifference/index.ts
```diff
@@ -47,29 +49,35 @@ const filters = [
 const isIntentionalDifference = (
   expected: AnnotatedHearingOutcome,
   actual: AnnotatedHearingOutcome,
-  incoming: AnnotatedHearingOutcome
+  incomingMessage: AnnotatedHearingOutcome | PncUpdateDataset
 ): boolean => {
-  const expectedMatch = summariseMatching(expected, true)
-  const actualMatch = summariseMatching(actual, true)
+  const intentionalDifference: IntentionalDifference = {
```
bjpirt, line 54: Naming-wise, this isn't an intentional difference, it's the comparison data we're using to determine if something is an intentional difference. So I wonder about naming this type something like `ComparisonData` to represent the intention?

bjpirt (approved):

packages/core/comparison/lib/isIntentionalDifference/index.ts
chubberlisk: Resolved in https://github.com/ministryofjustice/bichard7-next-core/pull/795/commits/6b449ed90988d66085f2346d362df6bd24d8e8ea.

bjpirt (approved):

PR #796, opened 2024-06-21
Implement task for more explicit locking in bichard phase 1

bjpirt:
Note: This hasn't been integrated into the workflow yet to avoid the workflows blocking until the task is deployed. Follow-up PR to come

How it works:
- S3 can have key:value tags on a file
- When we start processing a workflow, we check the file to make sure it hasn't already got a "locked" tag
- If it has we'll carry on and ignore it as it's already been processed
- If it hasn't we'll add one and continue

There is a small possibility of a race condition here, however whenever we've had multiple workflows being triggered they've been separated by ~30 seconds, so this window of a few milliseconds is extremely unlikely to impact us in our specific scenario

### DIFF
--- a/packages/core/conductor-tasks/common/lockS3File.ts
+++ b/packages/core/conductor-tasks/common/lockS3File.ts
@@ -1,0 +1,83 @@
+import type { ConductorWorker } from "@io-orkes/conductor-javascript"
+import completed from "@moj-bichard7/common/conductor/helpers/completed"
+import failed from "@moj-bichard7/common/conductor/helpers/failed"
+import failedTerminal from "@moj-bichard7/common/conductor/helpers/failedTerminal"
+import inputDataValidator from "@moj-bichard7/common/conductor/middleware/inputDataValidator"
+import type Task from "@moj-bichard7/common/conductor/types/Task"
+import createS3Config from "@moj-bichard7/common/s3/createS3Config"
+import readS3FileTags from "@moj-bichard7/common/s3/readS3FileTags"
+import writeS3FileTags from "@moj-bichard7/common/s3/writeS3FileTags"
+import { isError } from "@moj-bichard7/common/types/Result"
+import { z } from "zod"
+
+const s3Config = createS3Config()
+
+const inputDataSchema = z.object({
+  fileName: z.string(),
+  bucketId: z.string(),
+  workflowId: z.string()
+})
+type InputData = z.infer<typeof inputDataSchema>
+
+const { TASK_DATA_BUCKET_NAME } = process.env
+if (!TASK_DATA_BUCKET_NAME) {
+  throw Error("TASK_DATA_BUCKET_NAME environment variable is required")
+}
+
+const buckets: Record<string, string> = {
+  "task-data": TASK_DATA_BUCKET_NAME
+}
+
+const lockKey = "lockedByWorkstream"
+
+// This provides basic locking to try to make sure the workflow doesn't get triggered twice.
+// This implementation is not race-condition proof. However, the specific case we have
+// with duplicate workflows separated by a period of time will be prevented by this.
+// Currently there's a window of 2 - 5 seconds where multiple workflows could be started
+// for the same message - this reduces the window to a few milliseconds
+const lockS3File: ConductorWorker = {
+  taskDefName: "lock_s3_file",
+  execute: inputDataValidator(inputDataSchema, async (task: Task<InputData>) => {
+    const { fileName, bucketId, workflowId } = task.inputData
+
+    const bucket = buckets[bucketId]
+    if (!bucket) {
+      return failedTerminal(`Bucket ID was not found: ${bucketId}`)
+    }
+
+    const tags = await readS3FileTags(fileName, bucket, s3Config)
+    if (isError(tags)) {
+      // File has already been removed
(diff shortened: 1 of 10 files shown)


packages/core/conductor-tasks/common/lockS3File.ts
```diff
@@ -0,0 +1,83 @@
...
+  throw Error("TASK_DATA_BUCKET_NAME environment variable is required")
+}
+
+const buckets: Record<string, string> = {
+  "task-data": TASK_DATA_BUCKET_NAME
+}
+
+const lockKey = "lockedByWorkstream"
+
+const deleteS3File: ConductorWorker = {
```
chubberlisk, line 33: Copy and pasta - I think this should be `lockS3File`? 😝

bjpirt: Good spot, thanks!

packages/common/s3/readTagsFromS3.ts
chubberlisk, line 1: What do you think about being more explicit in the naming of these new functions?

- `readTagsInS3` to `readTagsforFileInS3`
-  `writeTagsToS3` to `writeTagsForFileInS3`

This is because you can have tags on a bucket and a file, and in these we're specifically talking about tags on a file.

packages/core/conductor-tasks/common/lockS3File.ts
```diff
@@ -0,0 +1,83 @@
...
+      }
+
+      // Error reading the tags
+      return failed(tags.message)
+    }
+
+    if (tags[lockKey]) {
+      // File has already been locked
+      return completed({ lockState: "failure" }, "S3 File already locked")
+    }
```
chubberlisk, line 62:
```suggestion
        return completed({ lockState: "alreadyDeleted" }, "S3 File already deleted")
      }

      // Error reading the tags
      return failed(tags.message)
    }

    if (tags[lockKey]) {
      // File has already been locked
      return completed({ lockState: "alreadyLocked" }, "S3 File already locked")
    }
```

Struggling with the words to explain this suggestion (sorry!), but there's something confusing around completing but then there reading failure here. Also, practically in the Conductor UI I think it's nicer to have all the information in the output data, instead of having to go to the logs to see what the "failure" is.

Does that make any sense? 😅

packages/core/conductor-tasks/common/lockS3File.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    const result = await writeTagsToS3(fileName, bucket, { [lockKey]: workflowId }, s3Config)
+    if (isError(result)) {
+      return failed(result.message)
+    }
+
+    const newTags = await readTagsFromS3(fileName, bucket, s3Config)
+    if (isError(newTags)) {
+      return failed(tags.message)
```
chubberlisk, line 71:
```suggestion
      return failed(newTags.message)
```

Should this be using `newTags` instead?

packages/core/conductor-tasks/common/lockS3File.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    if (!newTags[lockKey] || newTags[lockKey] !== workflowId) {
+      // File has already been locked
+      return failed("Error locking file")
+    }
```
chubberlisk, line 77: Don't think we have a test that covers this? 👀

packages/common/s3/readTagsFromS3.ts
bjpirt: Fair point - I've gone with
- `readS3FileTags`
- `writeS3FileTags`

packages/core/conductor-tasks/common/lockS3File.ts
bjpirt: Yep, makes sense - will see how I can make clearer

packages/core/conductor-tasks/common/lockS3File.ts
bjpirt: 👍

packages/core/conductor-tasks/common/lockS3File.ts
bjpirt: 🦅 👀

packages/core/conductor-tasks/common/lockS3File.ts
bjpirt: More thoughts on this - semantically, completed just means it didn't raise an unexpected error, so completing with a failed lockState is OK. This variable is going to be used in the workflow so it's easier to only have to deal with two states. In terms of seeing the failure, the `"S3 File already deleted"` and `"S3 File already locked"` messages will get put into the logs in Conductor so that will be easily understandable.

packages/core/conductor-tasks/common/lockS3File.ts
bjpirt: I don't think it's possible to test this - this is to reduce the chance of a race condition so we'd have to somehow change the tags between writing them and reading them in this task

chubberlisk: Ahhhhh yushhhh, that makes sense - not sure how we'd do that!

**_🦅 flies away..._**

chubberlisk (dismissed):
Thank you for addressing my comments and suggestions! ✅

PR #797, opened 2024-06-21
Add the lock task to Conductor

bjpirt:
Also - rename a variable for consistency

### DIFF
--- a/packages/conductor/tasks/lock_s3_file.json
+++ b/packages/conductor/tasks/lock_s3_file.json
@@ -1,0 +1,11 @@
+{
+  "name": "lock_s3_file",
+  "retryCount": 10,
+  "retryLogic": "EXPONENTIAL_BACKOFF",
+  "retryDelaySeconds": 10,
+  "timeoutSeconds": 30,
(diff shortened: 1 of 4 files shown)


packages/conductor/tasks/lock_s3_file.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "lock_s3_file",
+  "retryCount": 10,
+  "retryLogic": "EXPONENTIAL_BACKOFF",
+  "retryDelaySeconds": 10,
+  "timeoutSeconds": 30,
+  "responseTimeoutSeconds": 30,
+  "timeoutPolicy": "RETRY",
+  "inputKeys": ["fileName", "bucketId", "lockId"],
+  "outputKeys": []
```
chubberlisk, line 10: Don't we need to define `lockState` here as it's outputted when the task is completed?

bjpirt: doh!

chubberlisk (approved):
:shipit:

PR #803, opened 2024-06-24
Update the bichard_phase_1 workflow to use the s3 file locking task

bjpirt:
This wires the bichard_phase_1 workflow to use the task created in #797 - now we check to see if we can get a lock and if we can't or the file is already gone, we exit gracefully.

Should prevent the slack alerts complaining about failed workflows


C-gyorfi (approved):
LGTM

PR #807, opened 2024-06-24
Add Phase 2 intentional difference for Bichard incorrectly raising HO200114

chubberlisk:
## Context

We ran the comparison tests for Phase 2, and it highlighted a failure where Bichard raised a HO200114 exception and Core didn't. Upon investigating, we realised that this was a bug in Bichard (see comments in intentional difference function). 

- Update Phase 2 comparison test to check for intentional differences. This will also fix whitespace failures when running the tests.
- Add a new intentional difference function to cover the case when Bichard incorrectly raises a HO200114.

### DIFF
--- a/packages/core/comparison/lib/isIntentionalDifference/index.ts
+++ b/packages/core/comparison/lib/isIntentionalDifference/index.ts
@@ -26,6 +26,7 @@ import nonMatchingManualSequenceNumber from "./nonMatchingManualSequenceNumber"
 import offenceReasonSequenceFormat from "./offenceReasonSequenceFormat"
 import prioritiseNonFinal from "./prioritiseNonFinal"
 import trailingSpace from "./trailingSpace"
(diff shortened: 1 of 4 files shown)


packages/core/comparison/lib/isIntentionalDifference/index.ts
```diff
@@ -44,7 +45,8 @@ const filters = [
   invalidManualSequenceNumber,
   nonMatchingManualSequenceNumber,
   offenceReasonSequenceFormat,
-  prioritiseNonFinal
+  prioritiseNonFinal,
+  bichardRaisesHo200114ForNonExactSequenceNumbers
```
bjpirt, line 49: I think the rest of the items in this array are in alphabetical order

chubberlisk: Ahhhhh yupppp, they are - I've fixed that now!

bjpirt (approved):

PR #810, opened 2024-06-25
Fix HO200106 exception not being raised in handleSentence function

chubberlisk:
## Context

When running Phase 2 comparison tests, there were a couple of failures around Core not raising HO200106 exceptions, but Bichard did.

Upon investigation, we realised that our rewrite logic (`/lib/deriveOperationSequence/handleSentence` function) for `UpdateMessageSequenceBuilderImpl.java:1153` (the `deriveOperationSequence` function) didn't quite match up, and we were retuning early when adjudication don't exist.

Fix the `handleSentence` function to match Bichard, and update unit tests. We removed one as that was testing for the incorrect early return.

### DIFF
--- a/packages/core/phase2/lib/deriveOperationSequence/handleSentence.test.ts
+++ b/packages/core/phase2/lib/deriveOperationSequence/handleSentence.test.ts
@@ -251,9 +251,9 @@ describe("handleSentence", () => {
     )
   })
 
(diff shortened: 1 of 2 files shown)


bjpirt (approved):

chubberlisk:
I've run comparison tests covering random 2/3 days periods since records began for Phase 2 (2023-02-08), and seems all good so merging this in!

PR #830, opened 2024-06-28
More phase 2 refactoring

emadkaramad:
In this PR:
- removed force casting to `PncDisposal` in `createPncDisposal` function.
- updated `PncResultQuery` schema to make `type` and `sentenceDate` optional. In Phase 2, these values can be `undefined`.
- Simplified some of the Phase 2 regex expressions
- Refactored some of the Phase 2 functions

### DIFF
--- a/packages/core/phase1/schemas/pncQueryResult.ts
+++ b/packages/core/phase1/schemas/pncQueryResult.ts
@@ -8,12 +8,12 @@ export const pncDisposalSchema = z.object({
   qtyUnitsFined: z.string().optional(),
   qualifiers: z.string().optional(),
   text: z.string().optional(),
-  type: z.number()
+  type: z.number().optional()
 })
 
 export const pncAdjudicationSchema = z.object({
   verdict: z.string(),
-  sentenceDate: z.coerce.date(),
+  sentenceDate: z.coerce.date().optional(),
   offenceTICNumber: z.number(),
   plea: z.string(),
   weedFlag: z.string().optional()
--- a/packages/core/phase2/createPncDisposal.ts
+++ b/packages/core/phase2/createPncDisposal.ts
@@ -26,7 +26,7 @@ const createPncDisposal = (
   const qtyDuration = durationUnit ? durationUnit + durationLength?.toString() : ""
   return {
     qtyDuration: qtyDuration,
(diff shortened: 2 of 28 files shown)


packages/core/phase1/schemas/pncQueryResult.ts
```diff
@@ -8,12 +8,12 @@ export const pncDisposalSchema = z.object({
...
   qualifiers: z.string().optional(),
   text: z.string().optional(),
-  type: z.number()
+  type: z.number().optional()
 })
 
 export const pncAdjudicationSchema = z.object({
   verdict: z.string(),
-  sentenceDate: z.coerce.date(),
+  sentenceDate: z.coerce.date().optional(),
```
emadkaramad, line 16: Phase 1 comparison tests passed after making `sentenceDate` and `type` optional in the schema and updating Phase 1 code.

packages/core/phase2/createPncDisposal.ts
```diff
@@ -43,7 +43,7 @@ const createPncDisposal = (
     ),
     text: disposalText,
     type: pncDisposalType
-  } as PncDisposal
```
emadkaramad, line 46: The changes in the `PncQueryResult` schema are related to removing this type casting.
- qtyMonetaryValue is string in the schema, but we were assign a number value here.
- `sentenceDate` and `type` can be undefined in Phase 2 but are required fields in the schema. To bypass ts error we were force casting to `PncDisposal` type to bypass the ts error.

packages/core/phase2/createPncDisposal.ts
```diff
@@ -26,7 +26,7 @@ const createPncDisposal = (
   const qtyDuration = durationUnit ? durationUnit + durationLength?.toString() : ""
   return {
     qtyDuration: qtyDuration,
-    qtyMonetaryValue: amtSpecInResult ? amtSpecInResult : "",
+    qtyMonetaryValue: amtSpecInResult?.toString(),
```
emadkaramad, line 29: `qtyMonetaryValue` is a string type in the schema, but we assigned a number value here and force cast to `PncDisposal` to bypass ts error.

bjpirt (approved):

PR #849, opened 2024-07-11
Refactor of TRPS0013 test to improve coverage.

Senny10:
Improved test to cover the case when an offence is recordable and has a number of offence TICs that it should return trigger TRPS0013 and the court offence sequence number.

### DIFF
--- a/packages/core/phase1/triggers/TRPS0013.test.ts
+++ b/packages/core/phase1/triggers/TRPS0013.test.ts
@@ -92,19 +92,9 @@ describe("TRPS0013", () => {
 
   it.each([
     { isRecordable: false, hasCompletedDisarr: false },
-    { isRecordable: false, hasCompletedDisarr: false },
     { isRecordable: true, hasCompletedDisarr: false },
     { isRecordable: false, hasCompletedDisarr: true },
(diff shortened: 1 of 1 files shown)


packages/core/phase1/triggers/TRPS0013.test.ts
```diff
@@ -87,23 +91,23 @@ describe("TRPS0013", () => {
...
+    { isRecordable: false, hasCompletedDisarr: false },
+    { isRecordable: true, hasCompletedDisarr: false },
+    { isRecordable: true, hasCompletedDisarr: true },
+    { isRecordable: true, hasCompletedDisarr: true },
+    { isRecordable: false, hasCompletedDisarr: true },
+    { isRecordable: false, hasCompletedDisarr: false },
+    { isRecordable: false, hasCompletedDisarr: true },
+    { isRecordable: true, hasCompletedDisarr: false },
+    { isRecordable: true, hasCompletedDisarr: false },
+    { isRecordable: false, hasCompletedDisarr: true }
```
chubberlisk, line 107: This could be reduced to only 4 scenarios now that we've removed the redundant parameters.

PR #852, opened 2024-07-11
Write tests for TRPR0002

BushraAbdullahi:


packages/core/phase1/triggers/TRPR0002.test.ts
```diff
@@ -0,0 +1,54 @@
+import TriggerCode from "bichard7-next-data-latest/dist/types/TriggerCode"
+import generateAhoFromOffenceList from "../../phase2/tests/fixtures/helpers/generateAhoFromOffenceList"
+import type { Offence } from "../../types/AnnotatedHearingOutcome"
+import TRPR0002, { resultCodes } from "./TRPR0002"
+
+describe("TRPR0002", () => {
+  it.each(resultCodes)(
```
bjpirt, line 7: I feel a bit uneasy about this - if resultCodes changes to something wrong then the test will still pass. Whilst duplication is generally not a good thing, in this case I think it's good because it's kind of like a double-check for what you're testing.

BushraAbdullahi: yeah that makes sense. thanks for pointing it out

emadkaramad (approved):

packages/core/phase1/triggers/TRPR0002.test.ts
```diff
@@ -0,0 +1,54 @@
+import TriggerCode from "bichard7-next-data-latest/dist/types/TriggerCode"
+import generateAhoFromOffenceList from "../../phase2/tests/fixtures/helpers/generateAhoFromOffenceList"
+import type { Offence } from "../../types/AnnotatedHearingOutcome"
+import TRPR0002 from "./TRPR0002"
+
+describe("TRPR0002", () => {
+  it.each([4575, 4576, 4577, 4585, 4586])(
```
emadkaramad, line 7: Nitpick: To avoid duplicating this array in our tests here, we can define a variable in this file and use that instead.

```
const resultCodes = [4575, 4576, 4577, 4585, 4586]

...

it.each(resultCodes)(...)

it.each(resultCodes)(...)
```

bjpirt: 👍

PR #863, opened 2024-07-15
remove TRPR0009 

Harriet-Hall:
One unit test as function will only ever return []


packages/core/phase1/triggers/TRPR0009.test.ts
```diff
@@ -0,0 +1,11 @@
+import generateAhoFromOffenceList from "../../phase2/tests/fixtures/helpers/generateAhoFromOffenceList"
+import type { Offence } from "../../types/AnnotatedHearingOutcome"
+import TRPR0009 from "./TRPR0009"
+
+const generateHearingOutcome = generateAhoFromOffenceList([] as Offence[])
+
+describe("TRPR0009", () => {
+  it("should not return trigger is shouldRaiseTrigger is false", () => {
```
BushraAbdullahi, line 8: the description isn't quite clear to me. Is this supposed to read as "should not return a trigger as shouldRaiseTrigger is false" ?

bjpirt (approved):

PR #887, opened 2024-07-25
TRPR0018 Unit tests

BushraAbdullahi:


packages/core/lib/triggers/TRPR0018.ts
```diff
@@ -6,7 +6,7 @@ import type { TriggerGenerator } from "../../types/TriggerGenerator"
 
 const triggerCode = TriggerCode.TRPR0018
 
-const findMatchingPncOffence = (
+export const findMatchingPncOffence = (
```
emadkaramad, line 9: A couple of suggestions:
- Consider keeping `findMatchingPncOffence` and `datesAreDifferent` as private functions (not exported). They seem to be internal helpers for `TRPR0018`, so it's generally best practice not to export them solely for testing purposes.
- Let's focus on directly testing the `TRPR0018` function for the unit tests. This is the main exported function, and thorough testing of its behaviour will implicitly cover the functionality of the internal helper functions.

This way we can ensure we're validating the public function (exported function) while maintaining the encapsulation of the module's internal workings.

packages/core/lib/triggers/TRPR0018.test.ts
```diff
@@ -0,0 +1,139 @@
...
+    expect(result).toEqual([{ code: triggerCode, offenceSequenceNumber: courtOffenceSequenceNumber }])
+  })
+
+  it("Should not generate a trigger when both start and end dates match", () => {
+    const pncQuery = generateMockPncQuery(true, false, startDate, endDate)
+    const aho = generateMockAho(true, startDate, endDate)
+    const result = TRPR0018({ ...aho, PncQuery: { ...pncQuery } })
+    expect(result).toEqual([])
+  })
+  it("Should not generate a trigger when PNC query has no matching court or penalty cases", () => {
```
emadkaramad, line 126: nitpick: missing a new line between test blocks.

emadkaramad (dismissed):

emadkaramad (approved):

PR #917, opened 2024-08-14
Persist phase 2

emadkaramad:
### PR Summary:

- **Dynamic Schema Selection:** The `s3TaskDataFetcher` now accepts a function that returns the schema, enabling dynamic selection of the schema for parsing input data in Phase 2.

- **New `triggersGenerated` Field:** Added a boolean field `triggersGenerated` to `Phase2Result`. This field indicates whether the Phase 2 code has invoked the function to generate triggers, helping to determine if the triggers should be updated in the database. The presence of an empty `triggers` array alone cannot reliably indicate if the function was called.

- **Enhanced `updateErrorListRecord` Logic:** The method now sets the status based on the number of exceptions and the previous status:
  - If exceptions are present, the status is set to `unresolved`.
  - If no exceptions exist and the status was previously `null`, it remains `null`.
  - If there are no exceptions and the status was not `null`, it is set to `resolved`.

- **Date Parsing Fix:** Resolved an issue with converting date strings to date objects in `newremOperationSchema`.

- **Updated `getOperationSequence`:** Now returns the `IgnoredAlreadyOnPNC` event code if results are already present on PNC.

- **Improved `isPncDisposalMatch` Function:** Adjusted the function to correctly match null, undefined, and empty strings.

- **Refined `saveErrorListRecord`:** Updated the function to:
  - Update the record if the case is ignored.
  - Avoid updating triggers when `triggersGenerated` is false (Phase 2).

- **Phase 2 Conductor Workflow Update:** Modified the workflow to ensure data persistence when a case is ignored.


packages/common/conductor/middleware/s3TaskDataFetcher.ts
```diff
@@ -35,7 +35,10 @@ const formatErrorMessages = (schema: string, error: z.ZodError): string[] =>
     return `${schema} parse error. Schema mismatch`
   })
 
-const s3TaskDataFetcher = <T>(schema: z.ZodSchema, handler: Handler<TaskDataInputData<T>>): OriginalHandler => {
+const s3TaskDataFetcher = <T>(
+  schema: z.ZodSchema | ((taskData: unknown) => z.ZodSchema),
```
emadkaramad, line 39: `s3TaskDataFetcher` accepts both ZodSchema and a function that returns ZodSchema.

packages/core/conductor-tasks/bichard_phase_2/processPhase2.ts
```diff
@@ -32,12 +43,13 @@ const processPhase2: ConductorWorker = {
       return failed(`Could not put file to S3: ${s3TaskDataPath}`, s3PutResult.message)
     }
 
-    const hasTriggersOrExceptions =
-      ("triggers" in result && result.triggers.length > 0) ||
-      ("hearingOutcome" in result && result.outputMessage.Exceptions.length > 0)
+    const hasTriggersOrExceptionsOrIgnored =
+      result.triggersGenerated ||
+      result.outputMessage.Exceptions.length > 0 ||
+      result.resultType === Phase2ResultType.ignored
```
emadkaramad, line 50: We need to persist phase 2 result, when
- `triggersGenerated` is true. That means the function to generate triggers has been called in Phase 2.
- There are any exceptions
- `resultType` is `ignored`. We also update the record when the case is ignored as it is possible that the record is resubmitted to Phase 1, and the error status is `Submitted`. So if case is ignored in Phase 2, we need to update the error_status in the database.

packages/core/phase2/lib/getOperationSequence/areAllResultsAlreadyPresentOnPnc/isMatchToPncAdjAndDis/isPncDisposalMatch.ts
```diff
@@ -1,12 +1,14 @@
 import type { PncDisposal } from "../../../../../types/PncQueryResult"
 
+const areStringsEqual = (obj1: string | undefined, obj2: string | undefined) => (!obj1 && !obj2) || obj1 === obj2
```
emadkaramad, line 3: `null`, `undefined`, and empty string are all considered equivalent in this context.

packages/core/lib/database/saveErrorListRecord.ts
```diff
@@ -8,15 +11,20 @@ import insertErrorListRecord from "./insertErrorListRecord"
...
 import updateErrorListTriggers from "./updateErrorListTriggers"
-import type PhaseResult from "../../types/PhaseResult"
-import { getAnnotatedHearingOutcome } from "../../types/PhaseResult"
 
 const handleUpdate = async (db: Sql, recordId: number, result: PhaseResult): Promise<void> => {
   const aho = getAnnotatedHearingOutcome(result)
-  if (aho.Exceptions.length > 0) {
+  // If case is resubmitted to Phase 1 and case is ignored in Phase 2,
+  // we should update the record to set the correct error_status
+  if (aho.Exceptions.length > 0 || result.resultType === Phase2ResultType.ignored) {
```
chubberlisk, line 19: >If case is resubmitted to Phase 1

Is it the `aho.Exceptions.length > 0` that suggests it's resubmitted to Phase 1? If this is used for Phase 1 and 2 then it's not just if it's resubmitted to Phase 1 but also if Phase 2 generates an exception?

packages/core/phase1/tests/helpers/PartialPncUpdateDataset.ts
```diff
@@ -0,0 +1,5 @@
+import type { z } from "zod"
```
chubberlisk, line 1: Why does this live in Phase 1? 🤔

packages/core/phase1/tests/helpers/generateMockPhase2Result.ts
```diff
@@ -0,0 +1,26 @@
+import { randomUUID } from "crypto"
```
chubberlisk, line 1: Why does this live in Phase 1? 🤔

packages/core/conductor-tasks/bichard_phase_2/processPhase2.ts
chubberlisk: For additional context on `triggersGenerated` from our Slack convo:

>If we don't use `triggersGenerated` and only rely on the `triggers` array, then `updateErrorListTriggers` will delete existing triggers because it compares existing triggers with the `triggers` array and delete the ones that don't exist in the `triggers` array.

packages/conductor/workflows/bichard_phase_2.json
```diff
@@ -54,10 +54,10 @@
           "permissive": false
         },
         {
-          "name": "check_triggers_or_exceptions",
-          "taskReferenceName": "check_triggers_or_exceptions",
+          "name": "check_triggers_or_exceptions_or_ignored",
+          "taskReferenceName": "check_triggers_or_exceptions_or_ignored",
           "inputParameters": {
-            "hasTriggersOrExceptions": "${process_phase2.output.hasTriggersOrExceptions}"
+            "hasTriggersOrExceptions": "${process_phase2.output.hasTriggersOrExceptionsOrIgnored}"
```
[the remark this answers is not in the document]
emadkaramad: Thanks! Updated all namings.

packages/core/conductor-tasks/bichard_phase_2/persistPhase2.integration.test.ts
```diff
@@ -0,0 +1,291 @@
+jest.setTimeout(30_000)
```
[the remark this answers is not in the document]
emadkaramad: Tests added.

packages/core/lib/database/saveErrorListRecord.ts
emadkaramad: No, it's not the exceptions length that indicates if the case is resubmitted. We don't check explicitly if the case is resubmitted here. This is to ensure that we don't end up having a case in Submitted status.

I haven't checked the Phase 1 code, but even if the case is ignored in Phase 1 after resubmission to Phase 1, then this code should cover it as well.

packages/core/phase1/tests/helpers/generateMockPhase2Result.ts
emadkaramad: Same!

packages/core/conductor-tasks/bichard_phase_2/processPhase2.integration.test.ts
```diff
@@ -0,0 +1,135 @@
+import "../../phase1/tests/helpers/setEnvironmentVariables"
```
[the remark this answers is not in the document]
emadkaramad: Tests added

chubberlisk (dismissed):
Thank you for addressing my comments and questions! ✅

packages/core/conductor-tasks/bichard_phase_2/processPhase2.ts
```diff
@@ -8,21 +8,32 @@ import { AuditLogEventSource } from "@moj-bichard7/common/types/AuditLogEvent"
...
+import type { PncUpdateDataset } from "../../types/PncUpdateDataset"
+import { isPncUpdateDataset } from "../../types/PncUpdateDataset"
 
 const s3Config = createS3Config()
 const taskDataBucket = process.env.TASK_DATA_BUCKET_NAME ?? "conductor-task-data"
 const lockKey: string = "lockedByWorkstream"
 
+const schemaFn = (taskData: unknown) =>
+  isPncUpdateDataset(taskData as AnnotatedHearingOutcome) ? pncUpdateDatasetSchema : unvalidatedHearingOutcomeSchema
+
```
bjpirt, line 25: I have a feeling we should be able to do this without a function - if we created a new schema that was a union of the two schemas it might be able to parse

packages/core/lib/database/updateErrorListRecord.ts
```diff
@@ -16,7 +18,6 @@ const generateUpdateFields = (result: PhaseResult): Partial<ErrorListRecord> =>
     error_quality_checked: record.error_quality_checked,
     annotated_msg: record.annotated_msg,
     updated_msg: record.updated_msg,
-    error_status: ResolutionStatus.UNRESOLVED,
```
bjpirt, line 19: Where does the error_status get set now?

packages/core/conductor-tasks/bichard_phase_2/processPhase2.ts
emadkaramad: Yes, that works. Thanks! I've updated the code.
I added the new schema in `processPhase2.ts`, but let me know if you think moving it to a separate file would be better.

packages/core/lib/database/updateErrorListRecord.ts
emadkaramad:
```
      UPDATE br7own.error_list SET ${db(updateFields)},
        error_status = CASE
          WHEN ${exceptionsCount}::integer > 0 THEN ${ResolutionStatus.UNRESOLVED}::integer
          WHEN error_status IS NULL THEN NULL
          ELSE ${ResolutionStatus.RESOLVED}::integer
        END
      WHERE error_id = ${recordId}
```

bjpirt (approved):

PR #929, opened 2024-08-28
Update message forwarder to take in Conductor workflow name

chubberlisk:
## Context

We're looking to set up a message forwarder for Phase 2 resubmissions, however atm it's hardcoded to start a Phase 1 workflow.

- Update message forwarder to take in a Conductor workflow name via an environment variable called `CONDUCTOR_WORKFLOW`.
  - By default, this is set to `bichard_phase_1`.
  - We do a check for valid workflows like destination type.

https://dsdmoj.atlassian.net/browse/BICAWS7-2994

### DIFF
--- a/packages/message-forwarder/src/forwardMessage/forwardMessage.ts
+++ b/packages/message-forwarder/src/forwardMessage/forwardMessage.ts
@@ -11,6 +11,8 @@ enum DestinationType {
   CONDUCTOR = "conductor"
 }
 
+const conductorWorkflows = ["bichard_phase_1", "bichard_phase_2"]
+
 const forwardMessage = async (
   message: string,
   stompClient: Client,
@@ -19,6 +21,11 @@ const forwardMessage = async (
   const destinationType: DestinationType = (process.env.DESTINATION_TYPE ?? "auto") as DestinationType
   if (!Object.values(DestinationType).includes(destinationType)) {
     return new Error(`Unsupported destination type: "${destinationType}"`)
+  }
+
+  const conductorWorkflow = process.env.CONDUCTOR_WORKFLOW ?? "bichard_phase_1"
+  if (!conductorWorkflows.includes(conductorWorkflow)) {
+    return new Error(`Unsupported Conductor workflow: "${conductorWorkflow}"`)
   }
 
   const aho = parseAhoXml(message)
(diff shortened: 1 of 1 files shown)


packages/message-forwarder/src/forwardMessage/forwardMessage.ts
```diff
@@ -21,6 +23,11 @@ const forwardMessage = async (
     return new Error(`Unsupported destination type: "${destinationType}"`)
   }
 
+  const conductorWorkflow = process.env.CONDUCTOR_WORKFLOW ?? "bichard_phase_1"
+  if (!conductorWorkflows.includes(conductorWorkflow)) {
+    return new Error(`Unsupported Conductor workflow: "${conductorWorkflow}"`)
+  }
+
   const aho = parseAhoXml(message)
```
emadkaramad, line 31: Should we parse both Aho and PncUpdateDataset?

bjpirt: Good point

bjpirt: I think this is just so we can extract the correlation ID for logging, maybe we should pull out the correlation ID extraction into its own function?

chubberlisk: Oooooooo yeahhh, that makes sense - alright if I do that in a separate PR?

emadkaramad: What about line 54?

return startBichardProcess(conductorWorkflow, aho, correlationId, conductorClient)
Does it start the new workflow with an AHO as input?

emadkaramad: It's safe to merge this PR. I just want to make sure when we use this for Phase 2, we don't miss PNC operations in the json.

emadkaramad (approved):

packages/message-forwarder/src/forwardMessage/forwardMessage.ts
chubberlisk: Hmmm, you're right Emad - not just for correlation ID as `startBichardProcess` adds the message to S3.

chubberlisk: Thanks for the spot!

PR #948, opened 2024-09-11
Use an enum for PNC operation codes

BushraAbdullahi:


packages/core/lib/serialise/pncUpdateDatasetXml/serialiseToXml.ts
```diff
@@ -17,7 +18,7 @@ const mapOperationStatus = (status: OperationStatus): OperationStatusXml => {
 
 const mapOperationToXml = (pncOperations: Operation[]): PncOperationXml[] => {
   return pncOperations.map<PncOperationXml>((operation) => {
-    if (operation.code === "DISARR") {
+    if (operation.code === PncOperation.NORMAL_DISPOSAL) {
       return {
         operationCode: {
           DISARR: operation.data
```
chubberlisk, line 24:
```suggestion
          [PncOperation.NORMAL_DISPOSAL]: operation.data
```

I think we can replace the keys in this file with the enum as well.

chubberlisk: We didn't update the other keys with PNC operations in the file.

chubberlisk (approved):
:shipit:

PR #989, opened 2024-10-03
Add HO200121

BushraAbdullahi:


packages/core/phase2/exceptions/HO200121.ts
```diff
@@ -1,9 +1,22 @@
...
 import type { ExceptionGenerator } from "../../types/ExceptionGenerator"
+import isRecordableOffence from "../lib/isRecordableOffence"
 
-const generator: ExceptionGenerator = (_aho: AnnotatedHearingOutcome, _options): Exception[] => {
-  return []
+const generator: ExceptionGenerator = (aho: AnnotatedHearingOutcome): Exception[] => {
+  const exceptions: Exception[] = []
+  const offences = aho.AnnotatedHearingOutcome.HearingOutcome.Case.HearingDefendant.Offence
+
+  if (offences.filter(isRecordableOffence).length === 0) {
```
emadkaramad, line 12: This also can be implemented as:

```
return offences.filter(isRecordableOffence).length === 0 ? [{
      code: ExceptionCode.HO200121,
      path: errorPaths.case.asn
    }] : []
```

emadkaramad (approved):

emadkaramad (commented):
Do we also need to remove the code that generates the exception HO200121 from generateOperations.ts?

PR #990, opened 2024-10-04
remove HO200121 exception logic from generateOperations

BushraAbdullahi:


packages/core/phase2/lib/generateOperations/generateOperations.test.ts
```diff
@@ -182,63 +182,6 @@ describe("generateOperations", () => {
     ])
   })
 
-  it("generates HO200121 exception when there are no recordable offences", () => {
```
emadkaramad, line 185: I think we need to have a test for when there are no recordable offences. Something like:
`it returns no operations when there are no recordable offences`

packages/core/phase2/lib/generateOperations/generateOperations.test.ts
```diff
@@ -341,6 +284,31 @@ describe("generateOperations", () => {
...
+  it("Should return no operations when there are no recordable offences", () => {
+    mockedAreAllResultsOnPnc.mockReturnValue(false)
+    const aho = {
+      Exceptions: [],
+      AnnotatedHearingOutcome: {
+        HearingOutcome: {
+          Case: {
+            HearingDefendant: {
+              Offence: [
+                {
```
emadkaramad, line 296: This offence is recordable and the test is not covering getOperations.ts line 76. You can check it by adding a breakpoint on line 76 and running your test in debug mode.

To make this offence non-recordable, you can set `OffenceCategory` to `B7`. (see [isRecordableOffence.ts](https://github.com/ministryofjustice/bichard7-next-core/blob/17d9be116f6c45f24d5ae29818e1a9d16c67ac96/packages/core/phase2/lib/isRecordableOffence.ts#L7))

emadkaramad (approved):

PR #999, opened 2024-10-09
Add HO200124

BushraAbdullahi:


packages/core/phase2/exceptions/HO200124.ts
```diff
@@ -1,9 +1,44 @@
...
+
+  if (fixedPenalty || allResultsAlreadyOnPnc) {
+    return []
+  }
+
+  checkResultClassExceptions(aho, (offence, result, offenceIndex, resultIndex) => {
+    const ccrId = offence?.CourtCaseReferenceNumber || undefined
+
+    if (result.PNCAdjudicationExists) {
+      return []
```
emadkaramad, line 24:
```suggestion
      return
```

packages/core/phase2/lib/generateOperations/resultClassHandlers/handleJudgementWithFinalResult.test.ts
```diff
@@ -73,67 +73,6 @@ describe("handleJudgementWithFinalResult", () => {
     expect(operations).toStrictEqual([{ code: PncOperation.DISPOSAL_UPDATED, data: undefined, status: "NotAttempted" }])
   })
 
-  it("should only return exception HO200124 when HO200124 and HO200108 conditions are met", () => {
```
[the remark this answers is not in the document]
BushraAbdullahi: I changed it to when results are NOT on pnc because we want to test the logic of 
`  
if (!allResultsAlreadyOnPnc && hasUnmatchedPncOffences(aho, ccrId) && !offence.AddedByTheCourt) {
    return { operations: [], exceptions: [] }
  }
`

right? we want it to enter that block statement so allResultsAlreadyOnPnc must be false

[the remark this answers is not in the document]
emadkaramad: That's right. 👍

packages/core/phase2/lib/generateOperations/resultClassHandlers/handleJudgementWithFinalResult.test.ts
```diff
@@ -204,4 +146,14 @@ describe("handleJudgementWithFinalResult", () => {
     expect(exceptions).toHaveLength(0)
     expect(operations).toHaveLength(0)
   })
+
+  it("should return no operations when results are not on PNC, there are unmatched PNC offences, and the offence is not added by the court", () => {
```
emadkaramad, line 150: We also need the same test to `handleAdjournmentWithJudgement.test.ts`.

BushraAbdullahi: I took the test out as it was raising operations whether those conditions are met or not, so it was perhaps not relevant? Please let me know if my observation is incorrect

emadkaramad (approved):

PR #1050, opened 2024-10-25
Update addExceptionsToAho to handle multiple exceptions instead

chubberlisk:
https://dsdmoj.atlassian.net/browse/BICAWS7-2975

### DIFF
--- a/packages/core/phase1/phase1.ts
+++ b/packages/core/phase1/phase1.ts
@@ -61,9 +61,7 @@ const phase1 = async (
     resultType = Phase1ResultType.ignored
   } else {
     const exceptions = generateExceptions(enrichedHearingOutcome)
-    exceptions.forEach(({ code, path }) => {
-      addExceptionsToAho(enrichedHearingOutcome as AnnotatedHearingOutcome, code, path)
-    })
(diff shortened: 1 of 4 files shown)


ian-antking (approved):

packages/core/phase1/phase1.ts
```diff
@@ -61,9 +61,7 @@ const phase1 = async (
     resultType = Phase1ResultType.ignored
   } else {
     const exceptions = generateExceptions(enrichedHearingOutcome)
-    exceptions.forEach(({ code, path }) => {
-      addExceptionsToAho(enrichedHearingOutcome as AnnotatedHearingOutcome, code, path)
-    })
+    addExceptionsToAho(enrichedHearingOutcome as AnnotatedHearingOutcome, exceptions)
```
bjpirt, line 64: This looks better, though it made me notice the casting in this part of the code which is normally a bad sign - is there a reason it's there?

chubberlisk: I removed it and was able to build and run unit tests so seems like it's redundant.

bjpirt (approved):

PR #1054, opened 2024-10-28
Refactor Phase 2 and its code structure

chubberlisk:
- Refactor `phase2` function to remove `processMessage` function to make it easier to follow.
- Moving files closer to where it's used or to the root of the `lib` folder if it's shared in different folders.
- Add missing tests for `getCourtCaseReferenceFromOperation`.

https://dsdmoj.atlassian.net/browse/BICAWS7-2975


packages/core/phase2/lib/areAnyPncResults2007.test.ts
```diff
@@ -1,4 +1,4 @@
-import type { AnnotatedHearingOutcome, Offence } from "../../../types/AnnotatedHearingOutcome"
+import type { AnnotatedHearingOutcome, Offence } from "../../types/AnnotatedHearingOutcome"
 import areAnyPncResults2007 from "./areAnyPncResults2007"
```
richardcane, line 2: Is there a reason this can't be a more general function? `pncRecordHasDisposalType(2007)`

richrace: Good point.

chubberlisk: It looks like a disposal type with 2007 is something meaningful and the only thing we're interested in although I see your point. There is also `areAllPncResults2007` which could be made generic. I'll refactor this.

chubberlisk: Have opened https://github.com/ministryofjustice/bichard7-next-core/pull/1055 to address this feedback, thanks!

PR #1059, opened 2024-10-30
Bicaws7 3173 db config

Senny10:
We have database configs in Common and the UI. The aim is to have one base config. The UI has different options, so there will be some alternations.
Refactor of the  three configs:
- createDbConfig (in Common used by API and Core)
- PostgresHelper (in Common used by Core for tests)
- DatabaseConfig and getDataSource (in UI and used by UI)


packages/common/db/createDbConfig.ts
```diff
@@ -1,25 +1,21 @@
...
-  username: process.env.DB_USER ?? "bichard",
-  password: process.env.DB_PASSWORD ?? "password",
-  ssl: process.env.DB_SSL === "true" ? { rejectUnauthorized: false } : false,
-  max: 10,
-  idle_timeout: 20,
-  max_lifetime: 60 * 30,
-  onnotice: () => false
-})
+// Note: we are declaring both user and username properties
+// in the config object.This is because the pg-promise
```
richrace, line 4:
```suggestion
// in the config object. This is because the pg-promise
```

packages/common/db/baseConfig.ts
```diff
@@ -0,0 +1,19 @@
...
+  ssl: boolean
+  schema: string
+}
+
+export const baseConfig: DatabaseConfig = {
+  host: process.env.DB_HOST ?? process.env.DB_AUTH_HOST ?? "localhost",
+  user: process.env.DB_USER ?? process.env.DB_AUTH_USER ?? "bichard",
+  password: process.env.DB_PASSWORD ?? process.env.DB_AUTH_PASSWORD ?? "password",
+  database: process.env.DB_DATABASE ?? process.env.DB_AUTH_DATABASE ?? "bichard",
+  port: parseInt(process.env.DB_PORT ?? process.env.DB_AUTH_PORT ?? "5432", 10),
```
richrace, line 16:
```suggestion
  port: Number(process.env.DB_PORT ?? process.env.DB_AUTH_PORT ?? "5432"),
```

So we don't have to specify the base

packages/common/db/createDbConfig.ts
```diff
@@ -1,25 +1,21 @@
...
+
+const createDbConfig = () =>
+  ({
+    ...baseConfig,
+    username: baseConfig.user,
+    max: 10,
+    idle_timeout: 20,
+    max_lifetime: 60 * 30,
+    onnotice: () => false
+  }) as const
```
richrace, line 19: Why do we need the `as const`?

packages/ui/src/services/getDataSource.ts
```diff
@@ -9,32 +9,22 @@ import User from "./entities/User"
...
-    port: databaseConfig.port,
-    username: databaseConfig.user,
-    password: databaseConfig.password,
-    database: databaseConfig.database,
+    host: baseConfig.host,
+    port: baseConfig.port,
+    username: baseConfig.user,
+    password: baseConfig.password,
+    database: baseConfig.database,
+    ssl: baseConfig.ssl,
```
richrace, line 22: Can't we do what we did in the `common/db/createDbConfig`?

packages/ui/src/services/getDataSource.ts
```diff
@@ -9,32 +9,22 @@ import User from "./entities/User"
...
     synchronize: false, // It must be always false, otherwise it changes the database structure.
     entities: [CourtCase, User, Trigger, Note, SurveyFeedback],
     subscribers: [CourtCaseSubscriber, NoteSubscriber],
     migrations: [],
-    schema: databaseConfig.schema,
-    ssl: databaseConfig.ssl ? { rejectUnauthorized: false } : false,
+    schema: "br7own",
```
richrace, line 27: This is in the `baseConfig`

packages/common/db/createDbConfig.ts
Senny10: This was used as suggested by Copilot to ensure type safety. The object would conform exactly as expected

richrace (approved):

PR #1060, opened 2024-10-31
Add gateways to API

richrace:
We know we need to add Audit Logs and upload them to an S3 bucket for conductor. 

This is the first go at naming these gateways.

- Are these file names meaningful and show intention?
- Are the interfaces meaningful and show intention?
- Are the implementation class names meaningful and show intention (including the fakes)?


Senny10 (dismissed):

richardcane:
Rich and I have just decided that the logical names of the gateway implementations should probably have more semantic meaning to the domain, so `DataStoreGateway` would become something like `db` or `Postgres` because that means more to the real implementation, which will make things easier to read.

richardcane (approved):

PR #1062, opened 2024-11-01
Add lint step for API in CI

richrace:


chubberlisk (commented):
I don't think this is needed because we have [lint_and_build](https://github.com/ministryofjustice/bichard7-next-core/blob/9b500e272406537ab55cc72fbf8f6ebec0aea0c4/.circleci/continue_config.yml#L192-L203) CircleCI job which always runs and covers the API.

If you check [your CircleCI run for this PR](https://app.circleci.com/pipelines/github/ministryofjustice/bichard7-next-core/6360/workflows/98b9f09a-05b9-4638-9c97-597451930c0a/jobs/53542), it shows it ran ESLint for the API.

richrace:
Good spot. Forgot about that 😅

PR #1077, opened 2024-11-07
Use esbuild for API

richrace:
- No helpful Swagger docs in Docker/Prod
- Only helpful Swagger docs in the Dev environment
- ~701MB Docker image size


richrace:
TSC prefered to esbuild

PR #1093, opened 2024-11-14
Remove force-excluded triggers from trigger-groups on filter panel

tausifpatel:
When the triggers are excluded for a force, then those excluded triggers should not appear in the trigger groups on filter panel for a logged in user - making filter panel clutter free.

When the user has access to multiple forces then hide only user excluded triggers from filter panel


packages/ui/src/utils/triggerGroups/allExcludedTriggers.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  it("Returns array with only user excluded triggers when non of the triggers are excluded on force level", () => {
+    const forceExcludedTriggers = { "01": [] }
+
+    const excludedTriggers = allExcludedTriggers(testUser, forceExcludedTriggers)
+
+    expect(excludedTriggers).toHaveLength(1)
+    expect(excludedTriggers).toContain(TriggerCode.TRPR0001)
+  })
+
+  it("Returns array with only force excluded triggers when non of the triggers are excluded on user level", () => {
```
chubberlisk, line 31:
```suggestion
  it("Returns array with only force excluded triggers when no triggers are excluded on user level", () => {
```

packages/ui/cypress/e2e/case-list/filter-cases/triggersExcludedWithMultipleForces.cy.ts
```diff
@@ -0,0 +1,55 @@
+describe("Triggers excluded at force level and user level for a user with access to multiple forces", () => {
```
chubberlisk, line 1: A bit confused by this, looking at `allExcludedTriggers` it seems we should have a test where it only shows excluded triggers for the user if they have multiple visible forces?

packages/ui/cypress/fixtures/users.ts
```diff
@@ -134,16 +134,35 @@ const users: Record<string, Partial<User> & { groups: UserGroup[] }> = {
...
+  },
+  Bichard18: {
+    username: "Bichard18",
+    visibleForces: ["18"],
+    forenames: "Bichard18",
+    surname: "User",
+    email: "bichard18@example.com",
+    password: hashedPassword,
+    groups: [UserGroup.GeneralHandler, UserGroup.NewUI]
+  },
```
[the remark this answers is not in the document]
tausifpatel: The numberedUsers() function generates users with access to visible forces up to 5. We needed to assign a force to the user that has no triggers excluded at the force level, allowing us to display all trigger codes on the filter panel. Since Force 18 is the only one without excluded triggers, I created a user with Force 18 as their visible force.

packages/ui/src/utils/triggerGroups/allExcludedTriggers.test.ts
tausifpatel: Fixed the typo

packages/ui/src/utils/triggerGroups/allExcludedTriggers.ts
```diff
@@ -0,0 +1,20 @@
+import type { DisplayFullUser } from "types/display/Users"
+
+const allExcludedTriggers = (
+  currentUser: DisplayFullUser,
+  forceExcludedTriggers: Record<string, string[]>
+): string[] => {
+  let allExcludedTriggers: string[] = []
+
+  if (currentUser.visibleForces.length === 1) {
```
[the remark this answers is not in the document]
tausifpatel: Added a test covering user with access to multiple forces

packages/ui/cypress/e2e/case-list/filter-cases/triggersExcludedWithMultipleForces.cy.ts
tausifpatel: This test coverage verifies that we only exclude user-level triggers when a user has access to multiple forces. For example, when a test user has access to forces `001`, `002`, and `003`, we still render `TRPR0008` even though it's excluded for force `001` in the standing data. This is because `TRPR0008` isn't excluded at the user level.

packages/ui/cypress/fixtures/users.ts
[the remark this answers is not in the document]
chubberlisk: Ahhhh okay, makes sense - maybe we can rename it so it says it's like `userWithoutExcludedTriggers` or something then so it's clearer than `Bichard18`?

chubberlisk (approved):
🎉

PR #1101, opened 2024-11-18
Update PNCErrorMessage to be an array of strings

chubberlisk:
## Context

We've implemented the remand PNC update request generator in #1092 but didn't implement the generation of exceptions. These can include errors we can get back from the PNC.

A previous assumption from Phase 1 is that we'll only get one message. However upon inspecting Phase 3 message, we now know this isn't true.

- Update PNCErrorMessage to be an array of strings and make all the necessary changes.
- Update PNC gateways to return a `PncApiError` which can hold multiple messages.

https://dsdmoj.atlassian.net/browse/BICAWS7-3231


packages/core/comparison/lib/MockPncGateway.ts
```diff
@@ -2,21 +2,27 @@ import type { PromiseResult } from "@moj-bichard7/common/types/Result"
...
     return Promise.resolve(this.result)
   }
 
   update(request: PncUpdateRequest, _correlationId: string): PromiseResult<void> {
     this.updates.push(request)
-    return Promise.resolve()
+
+    if (this.result instanceof PncApiError) {
+      return Promise.resolve(this.result)
+    } else {
```
richardcane, line 24: Don't need else here

packages/core/lib/serialise/ahoXml/serialiseToXml.ts
```diff
@@ -495,6 +498,14 @@ const xmlnsTags = {
   "@_xmlns:xsi": "http://www.w3.org/2001/XMLSchema-instance"
 }
 
+const mapPncExceptionsToXml = (exceptions: Exception[]): Br7PncErrorMessageString[] | undefined => {
+  const pncExceptions = exceptions.filter((exception) => "message" in exception) as PncException[]
+
+  return pncExceptions.length > 0
```
richardcane, line 504: This would read more clearly as an if/return than it does as ternary

```typescript
if(x) {
  return y
}
return
```

richardcane (approved):

packages/core/lib/serialise/ahoXml/serialiseToXml.ts
chubberlisk: We've included this changes in PR #1114, thanks!

PR #1102, opened 2024-11-19
  Note preview for cases in the case list should include the note username

BushraAbdullahi:
image


packages/ui/cypress/e2e/case-list/notePreview.cy.ts
```diff
@@ -0,0 +1,42 @@
+import { loginAndVisit } from "../../support/helpers"
+
+describe("Display username in note preview of latest note", () => {
```
richardcane, line 3: I think this can just be Note Preview

PR #1105, opened 2024-11-19
Add eslint-plugin-perfectionist

richardcane:
Implements https://perfectionist.dev/configs/recommended-natural as the perfectionist way to sort and order various bits of our codebase, including things like function parameters, object properties and imports.

Note: object property sorting is disabled for core because it breaks a lot of the snapshot testing, due to schemas being reorganised.


packages/ui/package.json
```diff
@@ -122,6 +122,7 @@
     "eslint-plugin-jsx-a11y": "^6.10.2",
     "eslint-plugin-mocha": "^10.5.0",
     "eslint-plugin-no-only-tests": "^3.3.0",
+    "eslint-plugin-perfectionist": "^3.9.1",
```
richrace, line 125: This should be installed at the top level as you include it at the top level `eslintrc`?

richardcane: Yes

PR #1108, opened 2024-11-20
Notes should be displayed in the same order as old UI

BushraAbdullahi:
Notes are now displayed in descending date order

image


richrace:
Is this wrong? 

https://github.com/ministryofjustice/bichard7-next-core/blob/BICAWS7-3229-notes-should-be-displayed-in-the-same-order-as-old-ui/packages/ui/src/services/dto/courtCaseDto.ts#L27

PR #1113, opened 2024-11-21
Add eslint-plugin-perfectionist

richrace:
Implements https://perfectionist.dev/configs/recommended-natural as the perfectionist way to sort and order various bits of our codebase, including things like function parameters, object properties and imports.

Note: object property sorting is disabled for core because it breaks a lot of the snapshot testing, due to schemas being reorganised.


chubberlisk (approved):
:shipit:

richardcane:
Going to break this down and submit PRs for each package as it's a lot to take in right now

PR #1119, opened 2024-11-25
Refactor UI health check

richrace:
- Check that the NextJS app is up rather than the nginx
- Refactor responsibility for verifying production has correct secrets
  - Keep the health endpoint free from things it doesn't care about
  - Throw an error in config so the app won't start in prod if it has default env parameters


richrace:
Pulled apart by #1121 and #1120

PR #1125, opened 2024-11-26
Add eslint-plugin-perfectionist to the API package

richardcane:
Implements https://perfectionist.dev/configs/recommended-natural as a way to sort and order various bits of our codebase, including things like function parameters, object properties and imports.


.eslintrc.js
```diff
@@ -76,6 +72,15 @@ module.exports = {
         "@typescript-eslint/no-explicit-any": "off",
         "jest/no-standalone-expect": "off"
       }
+    },
+    {
+      extends: ["plugin:perfectionist/recommended-natural-legacy"],
+      files: "packages/+(api)/**/*",
```
chubberlisk, line 78: Just an understanding question: what does this the + and () mean in `+(api)` mean? Is it "only in the api directory" and then when we do the other packages it'll be like `+(api|core)` or something?

chubberlisk (approved):
Thank you for breaking down the original PR, this was much quicker and easier to review! ✅

.eslintrc.js
richardcane: https://www.digitalocean.com/community/tools/glob

richardcane: + is the one or more matcher

richardcane: For this PR I could have just used `packages/api/**/*` but I was messing with that to make sure the glob pattern was picking up the right packages so thought I'd leave it in as I'm going to be adding to it shortly anyway :D

chubberlisk: Okayyy, that makes sense - thanks (and also for the glob tool, didn't know that was a thing)!

PR #1152, opened 2024-12-02
Send message to the relevant queue

tausifpatel:
Issue:
Currently, When the case is resubmitted, message is always getting submitted to `phase1ResubmissionQueue` even if the case is in Phase-2

Fix:
when a message is resubmitted, it goes into the appropriate queue based on the case's current phase.
If it is in phase-1 then the message gets sent to `phase1ResubmissionQueue`
If it is in phase-2 then the message gets sent to `phase2ResubmissionQueue`

Additional fix:
- Audit log resubmission event
- Return error when case is submitted in phase-2


packages/ui/src/services/resubmitCourtCase.ts
```diff
@@ -1,10 +1,13 @@
...
+import EventCategory from "@moj-bichard7/common/types/EventCategory"
+import EventCode from "@moj-bichard7/common/types/EventCode"
+import getAuditLogEvent from "@moj-bichard7/core/lib/getAuditLogEvent"
 import serialiseToXml from "@moj-bichard7/core/lib/serialise/ahoXml/serialiseToXml"
 import type { AnnotatedHearingOutcome } from "@moj-bichard7/core/types/AnnotatedHearingOutcome"
+import { AUDIT_LOG_EVENT_SOURCE } from "config"
 import amendCourtCase from "services/amendCourtCase"
 import type User from "services/entities/User"
 import insertNotes from "services/insertNotes"
-import sendToQueue from "services/mq/sendToQueue"
```
richrace, line 7: Why have you removed this? 

It has introduced more code in this file and others.

tausifpatel: function sendToQueue did not have any logic, it was just calling gateway.execute with messageXml and QueueName. Additionally, it created a new StompitMqGateway(config) instance with each function call, instead of having it injected as a dependency from the top level where it could be reused without creating new instances

richrace (approved):

PR #1167, opened 2024-12-04
Retry failed lint step

richrace:


ian-antking (dismissed):

scripts/lint-ci.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+ATTEMPTS=2
+WORKSPACES=($(grep "\"packages/" package.json | sed -E 's/[",]//g' | tr -s ' ' '\n' | awk '{print $1}'))
```
emadkaramad, line 4: Nitpick: If we have jq in the CI environment:

```
WORKSPACES=$(cat package.json | jq -r ".workspaces[]")
```

emadkaramad (approved):

scripts/lint-ci.sh
richrace: I wasn't sure, so I went with the universal option. (Also, we need to have it installed on our machines)

emadkaramad: We use `jq` in our scripts (in particular infrastructure scripts), so I think it should be available in our environment. I noticed we use `jq` in CircleCI config in core as well.

But it's just for readability, otherwise grep/sed/awk combination works everywhere. 👍

richrace: I updated it. It is much more readable 😅

emadkaramad (approved):

emadkaramad:
🙌

PR #1178, opened 2024-12-06
Upgrade to eslint 9

tausifpatel:


packages/common/mq/ActiveMqHelper.ts
```diff
@@ -108,7 +108,6 @@ class ActiveMqHelper {
     let waiting = true
 
     while (waiting) {
-      // eslint-disable-next-line no-await-in-loop
```
richardcane, line 111: I think we need to understand why these comments have disappeared in droves and make sure we haven't missed anything here.

[`no-await-in-loop` hasn't been removed from eslint@9](https://eslint.org/docs/latest/rules/no-await-in-loop). With specific reference to this file, it's not included in any configuration in the new eslint config.

richrace: The new config doesn't turn this on, i.e. `"no-await-in-loop": "error"`.

We might be defining plugins and extends wrong.

richrace: This was found in airbnb config which we no longer use

richardcane (dismissed):

PR #1185, opened 2024-12-10
Add test for remandGenerator

emadkaramad:


chubberlisk (approved):

bjpirt (dismissed):

packages/core/phase3/lib/pncUpdateRequestGenerators/remandGenerator/generateCourtNameType.test.ts
```diff
@@ -0,0 +1,131 @@
+import generateCourtNameType from "./generateCourtNameType"
+
+describe("generateCourtNameType", () => {
```
chubberlisk, line 3: I struggled with reading these tests and understanding the output from Jest:

image

I believe tests should be able to document the code without me having to look at the actual code.

In this instance, it's not immediately clear what the differences between each test like why does a court code of 0000 and 9998 make a difference? Of course, it turns out that 9998 is a value of `PNC_COURT_CODE_WHEN_DEFENDANT_FAILED_TO_APPEAR` so maybe use the constant or split these test cases into some contextual describe blocks e.g. `when the defendant failed to appear but doesn't have the failed to appear text`?

packages/core/phase3/lib/pncUpdateRequestGenerators/remandGenerator/getResultsForRemand.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    (resultClass) => {
+      const offences = [createOffence({ pncAdjudicationExists: true, resultClass })] as Offence[]
+
+      const results = getResultsForRemand(offences, operation)
+
+      expect(results).toHaveLength(0)
+    }
+  )
+
+  it("should return the matching result", () => {
```
chubberlisk, line 111: Is it possible to make one of the results non-matching? As all the results match meaning we're kind of just returning the results from each offence.

chubberlisk: Would also be good to extract some variables for the expectation so it's clear which results are being returned from which offence. I struggled to tell what was being matched and from where.

packages/core/phase3/lib/pncUpdateRequestGenerators/remandGenerator/getResultsForRemand.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const noAdjudicationResultClasses = [
+  ResultClass.ADJOURNMENT,
+  ResultClass.ADJOURNMENT_PRE_JUDGEMENT,
+  ResultClass.ADJOURNMENT_WITH_JUDGEMENT
+]
+
+const adjudicationResultClasses = [ResultClass.ADJOURNMENT_POST_JUDGEMENT, ResultClass.ADJOURNMENT_WITH_JUDGEMENT]
+
+const IsRemandOperationMatchingResult = (result: Result, operation: Operation<PncOperation.REMAND>) => {
```
chubberlisk, line 18: Think we want this to begin with a lowercase "i", `isRemandOperationMatchingResult`?

packages/core/phase3/lib/pncUpdateRequestGenerators/remandGenerator/remandGenerator.test.ts
```diff
@@ -0,0 +1,217 @@
...
+        courtNameType2: "Magistrates' Courts London Croydon MCA",
+        localAuthorityCode: "0000",
+        bailConditions: ["This is a dummy bail condition."]
+      }
+    })
+  })
+
+  it("should set bail conditions to empty array when PNC remand status is 'C'", () => {
+    const pncUpdateDataset = createPncUpdateDataset()
+    pncUpdateDataset.AnnotatedHearingOutcome.HearingOutcome.Case.HearingDefendant.RemandStatus = "PB"
```
chubberlisk, line 96: The test description mentions remand status is `C` but we set it to `PB`?

packages/core/phase3/lib/pncUpdateRequestGenerators/remandGenerator/remandGenerator.test.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    const result = remandGenerator(pncUpdateDataset, operation)
+
+    expect(result).toStrictEqual({
```
chubberlisk, line 100: I think we just do:

```suggestion
    expect(result.bailConditions).toHaveLength(0)
```

as we have already tested all the fields in the previous test so it's just noise in my opinion unless there are some other attributes that change as a result of changing the remand status.

packages/core/phase3/lib/pncUpdateRequestGenerators/remandGenerator/remandGenerator.test.ts
```diff
@@ -0,0 +1,217 @@
...
+        psaCourtCode: "9998",
+        courtNameType1: "Magistrates' Courts London Croydon MCA",
+        courtNameType2: "Magistrates' Courts London Croydon MCA",
+        localAuthorityCode: "0000",
+        bailConditions: []
+      }
+    })
+  })
+
+  it("should set court house code to 1st instance warrant issued when there is a result qualifier 'LE' and result code is undated warrant", () => {
```
chubberlisk, line 121: Again, with these tests below, maybe we just test what changes?

Could also do something like:

```
expect(result).toStrictEqual({
  operation: "NEWREM",
  request: expect.objectContaining({
    courtNameType1: "*****1ST INSTANCE DATED WARRANT ISSUED*****",
    courtNameType2: "*****1ST INSTANCE DATED WARRANT ISSUED***** MCA"
  })
})
```

chubberlisk (commented):
Much needed test coverage, thank you! I've added some comments/suggestions around test structure though.

packages/core/phase3/lib/pncUpdateRequestGenerators/remandGenerator/getResultsForRemand.test.ts
emadkaramad: Good shout. Updated. One is not unmatching and it checks the object references for the other 2 results.

packages/core/phase3/lib/pncUpdateRequestGenerators/remandGenerator/remandGenerator.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    const result = remandGenerator(pncUpdateDataset, operation)
+
+    expect(result).toStrictEqual({
+      operation: "NEWREM",
+      request: expectedRequest
+    })
+  })
+
+  it.each(["PB", "PR"])(
+    "should not retuen any bail conditions when PNC remand status is custody (%s)",
```
chubberlisk, line 97:
```suggestion
    "should not return any bail conditions when PNC remand status is custody (%s)",
```

chubberlisk (approved):
Thank you so much for addressing my feedback, the readability of the tests are massively improved and clearer to me! ❤️

chubberlisk (approved):

PR #1194, opened 2024-12-13
Remove feature flags

tausifpatel:
- Removed `exceptionsEnabled` feature flag
- Removed `offenceMatchingEnabled` feature flag


richardcane (approved):
I'm happy with this, though do just check that removing seed scripts doesn't cause any issues with liquibase.

PR #1197, opened 2024-12-17
Update CSRF token with interaction with React 

richrace:
- Add endpoint to get new CSRF token
- AutoSave updates the CSRF token
- Add a hook to call the CSRF endpoint
- Add refresh CSRF to case details tabs
- Add refresh CSRF to trigger exceptions panel
- Add refresh CSRF to offence list


packages/ui/src/pages/api/refresh-csrf-token.ts
```diff
@@ -0,0 +1,19 @@
...
+import generateCsrfToken from "middleware/withCsrf/generateCsrfToken"
+import type { NextApiRequest, NextApiResponse } from "next"
+
+export default async (request: NextApiRequest, response: NextApiResponse) => {
+  const allowedMethods = ["GET"]
+
+  const auth = await withApiAuthentication(request, response, allowedMethods)
+
+  if (!auth) {
+    return
```
emadkaramad, line 11: Should we set the status code to 401 here before returning?

emadkaramad (approved):

richardcane (approved):

PR #1226, opened 2025-01-08
API parse hearing outcome

richrace:
- Updated tests
- Copied parsing function from UI (we will remove this from UI in the future)
- Update zod schemas
- Use the fastify logger via dependency injection
- Added fixtures (should be removed from UI in the future)
- Added test helper for AHO files

Note: We have to build common without the Case type because the zod schema `unvalidatedHearingOutcomeSchema` is unavailable. It's not available because Core has not been built yet. So, we build the single file after it is available.


packages/api/src/services/parseHearingOutcome.ts
richardcane, line 1: Should this definitely live in the API?

packages/api/src/useCases/dto/convertCaseDBToDTO.ts
richardcane, line 1: Did we decide to keep `CaseDB` over `CaseRow`?

packages/common/package.json
```diff
@@ -5,8 +5,9 @@
   "main": "index.js",
   "scripts": {
     "build": "tsc --project tsconfig.build.json",
+    "build:api": "tsc --project tsconfig.build.api.json",
     "clean": "rm -rf ./dist",
-    "watch": "tsc --project tsconfig.build.json --watch",
+    "watch": "tsc --project tsconfig.build.api.json --watch",
```
emadkaramad, line 10: Should the command name be `watch:api`?

packages/api/src/services/parseHearingOutcome.ts
```diff
@@ -0,0 +1,36 @@
+import type { AnnotatedHearingOutcome } from "@moj-bichard7/core/types/AnnotatedHearingOutcome"
+import type { FastifyBaseLogger } from "fastify"
+
+import { isError } from "@moj-bichard7/common/types/Result"
+import parseAhoXml from "@moj-bichard7/core/lib/parse/parseAhoXml/parseAhoXml"
+import parseAnnotatedPncUpdateDatasetXml from "@moj-bichard7/core/phase2/parse/parseAnnotatedPncUpdateDatasetXml/parseAnnotatedPncUpdateDatasetXml"
+
+const isPncUpdateDatasetFromXml = (message: string) => /<AnnotatedPNCUpdateDataset/.exec(message)
+
+const parseHearingOutcome = (hearingOutcome: string, logger?: FastifyBaseLogger): AnnotatedHearingOutcome | Error => {
```
emadkaramad, line 10: I wonder if `logger` should be a required parameter. I would assume we always want to log errors?

packages/api/src/services/parseHearingOutcome.ts
emadkaramad: Only UI package needs this function it seems.

packages/api/src/services/parseHearingOutcome.ts
richardcane: I think I'd expect parseHearingOutcome to return a result or an error and the handler to do the logging

richrace: @emadkaramad We don't in tests.

@richardcane We use the logger instead of using `console.error` https://github.com/ministryofjustice/bichard7-next-core/blob/main/packages/ui/src/utils/parseHearingOutcome.ts#L24

packages/common/package.json
richrace: I'll make that change.

packages/api/src/useCases/dto/convertCaseDBToDTO.ts
richrace: We're having ongoing discussions about names.

richrace: https://dsdmoj.atlassian.net/browse/BICAWS7-3301

emadkaramad (approved):

packages/api/src/services/parseHearingOutcome.ts
richrace: @emadkaramad https://dsdmoj.atlassian.net/browse/BICAWS7-3311

PR #1239, opened 2025-01-13
PNCID should always be set from the PNC response

richardcane:
Currently, the PNCID won't be set if a case has exceptions. This PR changes the behaviour to always set the PNCID.
Also adds the court PNCID to the new UI.


packages/core/phase1/enrichAho/enrichFunctions/enrichWithPncQuery.ts
```diff
@@ -47,19 +47,18 @@ const clearPNCPopulatedElements = (aho: AnnotatedHearingOutcome): void => {
 }
 
 export default async (
-  annotatedHearingOutcome: AnnotatedHearingOutcome,
```
bjpirt, line 50: I think we were trying to avoid acronyms

packages/core/phase1/enrichAho/enrichFunctions/enrichWithPncQuery.ts
```diff
@@ -70,33 +69,37 @@ export default async (
...
-    annotatedHearingOutcome.AnnotatedHearingOutcome.HearingOutcome.Case.RecordableOnPNCindicator = true
+  if (aho.PncQuery !== undefined) {
+    aho.AnnotatedHearingOutcome.HearingOutcome.Case.RecordableOnPNCindicator = true
   }
 
-  annotatedHearingOutcome = matchOffencesToPnc(annotatedHearingOutcome)
+  aho = matchOffencesToPnc(aho)
 
-  return annotatedHearingOutcome
+  const { PNCIdentifier, CourtPNCIdentifier } = aho.AnnotatedHearingOutcome.HearingOutcome.Case.HearingDefendant
```
bjpirt, line 99: How sure are we there are no side effects from doing this? I don't think old Bichard did this as the comparison tests were passing so we need to be sure of the impact

bjpirt: Or is this just moved from `matchOffencesToPnc`?

packages/core/schemas/pncQueryResult.ts
```diff
@@ -52,7 +52,7 @@ export const pncQueryResultSchema = z.object({
   forceStationCode: z.string().describe(ahoDescription.AnnotatedHearingOutcome.PncQuery.forceStationCode.$description),
   croNumber: z.string().optional(),
   checkName: z.string(),
-  pncId: z.string(),
+  pncId: z.string().optional(),
```
bjpirt, line 55: Wouldn't mind seeing the example where this was not included

packages/core/phase1/enrichAho/enrichFunctions/enrichWithPncQuery.ts
richardcane: I'll put this back

packages/core/schemas/pncQueryResult.ts
richardcane: Think this has been a misconception on my part. Will remove.

packages/core/phase1/enrichAho/enrichFunctions/enrichWithPncQuery.ts
richardcane: Ben and I discussed this on a call - the old system was designed not to set these fields if there were exceptions, which is something we copied into core. This change causes these fields to be set all the time, which is something that's been queried in feedback from users.

We don't believe it will have any negative impact.

bjpirt (approved):

PR #1253, opened 2025-01-20
Lock exceptions and audit log

tausifpatel:
- [x] Add SQL to lock exceptions to user
- [x] Add useCase to add logic for locking
- [x] Add AuditLogging
- [x] Add transaction to useCase   
- [x] Rename `fetchCaseDto`
- [x] Create a test for the `lockAndFetchCase` use case
- [x] Create a new file for logic in the transaction, making it easier to test
- [x] Extract out exception locking into a new file
- [x] Add tests unit tests for existing code
- [x] Create new e2e tests for the `case/:caseId` endpoint with locking in mind


packages/api/src/useCases/cases/lockAndFetchCase.ts
```diff
@@ -0,0 +1,39 @@
...
+
+      const caseMessageId = (await dataStore.selectCaseMessageId(caseId)).message_id
+
+      await lockExceptions(dataStore, callbackSql, caseId, user, auditLogEvents)
+
+      // TODO: Check permissions for Triggers
+      // TODO: Lock Triggers and AuditLog them
+
+      if (auditLogEvents.length > 0) {
+        createAuditLogEvents(auditLogEvents, caseMessageId, auditLogGateway, logger)
```
bjpirt, line 31: This can return an error, but it doesn't throw. What happens to the transaction if there is an error creating an audit log event?

richrace: It's the same as in the UI (I think). 

The transaction only applies to the DB. I couldn't get the transaction to rollback with an error thrown in this function.

richrace: I missed the `await` 🤦‍♂️

PR #1262, opened 2025-01-22
Initial implementation of create audit log record endpoint

bjpirt:


packages/api/package.json
richrace, line 1: Can add the new dependencies in here like `axois`

richrace (approved):

PR #1263, opened 2025-01-23
Bichard cli

richardcane:
This P.R gives us a CLI tool to help with day-to-day tasks on Bichard. We are using a library called `Commander.js`
image

The current functionality covered in this tool includes:

- Run the elevate dev security groups 
- Fetch docker images from ECR
- Check PNC connectivity 
- Stop/start message processing 
- Open conductor - this fetches credentials from AWS - Kind of works with firefox you need to refresh the page for it to load correctly 
- Query cloudwatch for user login activity in User service and Postfix


packages/cli/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+Options:
+  -h, --help  display help for command
+
+```
+
+## Querying cloudwatch
+
+A user query that comes up time and time again is that a user can log into Bichard. We would previously have to log into AWS and check container logs for user activity. On most occasions the user would of provided the incorrect email that we have stored in Postgres.
+Now we can easily query our User service and postfix logs with a users email.
```
chubberlisk, line 50:
```suggestion
A common user query is that a user cannot log into Bichard. This is typically because the user has used an email address that doesn't match what we have stored in the database.

Previously, we had to log into AWS and check container logs for user activity but now we can easily query our User service and postfix logs with a users email.
```

packages/cli/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+`$ b7 cloudwatch user-service --start-time <hours> <email address>`
```
chubberlisk, line 52: Prefer multiline code blocks in these scenario because it provides you with the quick copy to clipboard button which makes it easier and quicker to copy the command and run.

````suggestion
```bash
b7 cloudwatch user-service --start-time <hours> <email address>
```
````

File: packages/cli/README.md [hunk 3]
```diff
@@ -0,0 +1,65 @@
...
+
+If you do not proide the `--start-time or -s` flag it defaults to an hour.
```
chubberlisk, line 54:
```suggestion
If you do not provide the `--start-time or -s` flag it defaults to an hour.
```

packages/cli/README.md
```diff
@@ -0,0 +1,65 @@
...
+The email address uses a wildcard so you can specify a partial email if we want to check for all user login for a specific time frame if users are reporting Bichard is down.
```
chubberlisk, line 55:
```suggestion
The email address uses a wildcard so you can specify a partial email if you want to check for all user logins for a specific time frame when users are reporting Bichard is down.
```

packages/cli/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+`b7 cloudwatch user-service --start-time 2 avonandsomerset`
```
chubberlisk, line 57:
````suggestion
```bash
b7 cloudwatch user-service --start-time 2 avonandsomerset
```
````

File: packages/cli/README.md [hunk 6]
```diff
@@ -0,0 +1,65 @@
...
+
+The above example would search for anyone with `avonandsomerset` in the past 2 hours
+
+## Conductor
+
+The command we have for opening Conductor fetches credentials from AWS and logs us into the Conductor UI
+
+> [!NOTE] We are using osascript to interact with the browser, the support for firefox is somewhat limited so and firefox users will need to refresh the browser on load.
```
chubberlisk, line 65: This doesn't render properly, think it needs to be like this:

```suggestion
> [!NOTE]
> We are using osascript to interact with the browser, the support for firefox is somewhat limited so and firefox users will need to refresh the browser on load.
```

chubberlisk (approved):
:shipit:

PR #1264, opened 2025-01-23
Port across 'create audit log events' code

bjpirt:


packages/api/package.json
```diff
@@ -65,6 +67,8 @@
     "http-status": "^1.8.1",
     "jsonwebtoken": "^9.0.2",
     "lodash": "^4.17.21",
```
richrace, line 69: We have the whole lodash library

richrace (approved):

PR #1270, opened 2025-01-24
Implement query audit logs and get one audit log

bjpirt:


packages/api/src/useCases/fetchAuditLogs/FetchAll.test.ts
richrace, line 1: Needs to be `FetchAll.unit.test.ts`

Same for other tests.

PR #1277, opened 2025-01-28
SPIKE: Make important information sticky

richrace:
https://github.com/user-attachments/assets/4898b0f0-c94f-46d0-b046-74a360fcdd9d


richrace:
Work is being done in https://github.com/ministryofjustice/bichard7-next-core/tree/BICAWS7-3320-case-details-defendant-details-header-design-update

PR #1290, opened 2025-02-03
Extract updatePnc function into its own file

chubberlisk:
https://dsdmoj.atlassian.net/browse/BICAWS7-3332

### DIFF
--- a/packages/core/phase3/lib/updatePnc.ts
+++ b/packages/core/phase3/lib/updatePnc.ts
@@ -1,3 +1,5 @@
+import { isError } from "@moj-bichard7/e2e-tests/utils/isError"
+
 import type PncGatewayInterface from "../../types/PncGatewayInterface"
 import type { PncUpdateDataset } from "../../types/PncUpdateDataset"
 import type PncUpdateRequest from "../types/PncUpdateRequest"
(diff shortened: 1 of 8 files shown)


packages/core/phase3/lib/updatePnc.ts
```diff
@@ -0,0 +1,42 @@
...
+  pncUpdateDataset: PncUpdateDataset,
+  pncUpdateRequest: PncUpdateRequest,
+  pncGateway: PncGatewayInterface
+) => {
+  const correlationId = pncUpdateDataset.AnnotatedHearingOutcome.HearingOutcome.Hearing.SourceReference.UniqueID
+
+  for (let pncLockErrorRetries = 0; pncLockErrorRetries < MAXIMUM_PNC_LOCK_ERROR_RETRIES; pncLockErrorRetries++) {
+    const pncUpdateResult = await pncGateway.update(pncUpdateRequest, correlationId)
+
+    if (isError(pncUpdateResult)) {
```
richardcane, line 26: Return early?

chubberlisk: Refactored so we return early to remove nested ifs - thank you for the suggestion!

richardcane (approved):

PR #1306, opened 2025-02-12
Remove conditional rendering for Add Note component

richardcane:
This PR removes the locking constraint applied to the Add Note component on the case details page. All users who have access to a case will now be able to add notes whether the case is resolved or not.


Senny10 (dismissed):

packages/ui/cypress/e2e/case-details/view-case-details/view-notes.cy.ts
```diff
@@ -133,18 +133,6 @@ describe("View notes", () => {
...
-      caseNotes: [[]],
-      force: "01"
-    })
-
-    loginAndGoToNotes()
-    cy.contains("Case has no notes.")
-    cy.get("textarea[name=noteText]").should("not.exist")
-    cy.get("#add-note-button").should("not.exist")
-  })
-
```
richrace, line 147: Should this test now be doing the opposite? We want to see the button now. Does another test cover this?

richrace: Also, that's a very specific cypress task, can it be removed if nothing else is using that?

richardcane: I'm going to remove any notion of the idea of "locking" in the tests as it's no longer relevant, I'll add another test for the adding notes to resolved cases requirement

richrace (dismissed):

richrace (approved):

PR #1309, opened 2025-02-14
Add api client

tausifpatel:


packages/ui/src/services/api/ApiClient.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+class ApiClient implements HttpClient {
+  readonly jwt: string
+
+  constructor(jwt: string) {
+    this.jwt = jwt
+  }
+
+  async get(route: string): Promise<unknown | Error> {
```
richrace, line 15:
```suggestion
  async get(route: string): Promise<unknown> {
```

Get rid of SonarCloud issues

packages/ui/src/services/api/interfaces/HttpClient.ts
```diff
@@ -0,0 +1,6 @@
+interface HttpClient {
+  readonly jwt: string
+  get: (route: string) => Promise<unknown | Error>
```
richrace, line 3:
```suggestion
  get: (route: string) => Promise<unknown>
```

richrace (approved):

PR #1318, opened 2025-02-20
Bicaws7 3320 case details header design update

Senny10:
The case details page has too much white space at the top. The case header has information that operators often refer back to, which means scrolling back to it all the time

The new design incorporates the action link buttons and the defendant's name into the header. The header will also be fixed to the top of the viewport when the user scrolls past the top of the component at its initially rendered location.

- Add the defendant's name and action link buttons to the header
- Fix the header to the top of the viewport when the user scrolls past it
- The header should always be visible
- The header should be anchored to the top of the viewport (possibly with padding, test with UCD)
- If the user scrolls back to top, header should return to its original position

Header details should still be accessible via keyboard at their original position in the tab navigation hierarchy

https://github.com/user-attachments/assets/0c0ad830-31fa-41c2-94a5-1553ce53ec45


packages/ui/src/features/CourtCaseDetails/Header.styles.tsx
```diff
@@ -1,17 +1,48 @@
...
+import { gdsLightGrey, grey } from "utils/colours"
+
+const CaseDetailHeaderContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  background-color: ${gdsLightGrey};
+  position: sticky;
+  top: 0;
+  z-index: 9;
+  padding: 10px 10px 10px 10px;
```
richrace, line 14: You have used `rem` above. Do you think you should use them here?

packages/ui/src/features/CourtCaseDetails/Header.styles.tsx
```diff
@@ -1,17 +1,48 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  gap: 30px;
+  width: 100%;
+
+  @media (max-width: 1300px) {
+    flex-direction: column;
+    align-items: flex-start;
+    gap: 0px;
```
richrace, line 26: Same here

packages/ui/src/features/CourtCaseDetails/Header.styles.tsx
```diff
@@ -1,17 +1,48 @@
...
+    }
+  }
+`
 
 const ButtonContainer = styled.div`
   display: flex;
   justify-content: flex-end;
-  margin-bottom: 24px;
+  margin-bottom: 0px;
+  padding-top: 5px;
```
richrace, line 38: And here

PR #1333, opened 2025-02-25
Add configuration to run e2e tests against api

tausifpatel:


packages/api/src/services/gateways/dataStoreGateways/postgres/visibleCourtsSql.ts
richrace, line 1: I had a chat with Ben about this so please ignore this code. It will be changed.

.circleci/continue_config.yml
```diff
@@ -830,6 +839,13 @@ jobs:
     parameters:
       ms-edge:
         type: boolean
+        default: false
+      api:
+        type: boolean
+        default: false
+      case-details:
```
chubberlisk, line 846: This can be vague, wondering if this should be `api-case-endpoint` like the other parameters in other jobs?

chubberlisk (approved):

.circleci/continue_config.yml
richrace: @tausifpatel is working on another CircleCI ticket, I'll make a note in the ticket 👍

PR #1340, opened 2025-03-04
API add filter defendant name

richrace:
We can now filter by the defendant's name

### DIFF
--- a/packages/api/src/useCases/cases/fetchCasesAndFlitersTests/filters/defendantName.e2e.test.ts
+++ b/packages/api/src/useCases/cases/fetchCasesAndFlitersTests/filters/defendantName.e2e.test.ts
@@ -1,0 +1,106 @@
+import type { User } from "@moj-bichard7/common/types/User"
+import type { FastifyInstance } from "fastify"
+
+import type { Pagination } from "../../../../types/CaseIndexQuerystring"
+
+import { createCases } from "../../../../tests/helpers/caseHelper"
+import { SetupAppEnd2EndHelper } from "../../../../tests/helpers/setupAppEnd2EndHelper"
+import { createUsers } from "../../../../tests/helpers/userHelper"
+import { fetchCasesAndFilter } from "../../fetchCasesAndFilter"
+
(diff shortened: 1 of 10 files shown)


packages/api/src/useCases/cases/fetchCasesAndFlitersTests/filters/defendantName.e2e.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    expect(caseMetadata.cases).toHaveLength(1)
+    expect(caseMetadata.totalCases).toBe(1)
+    expect(caseMetadata.returnCases).toBe(1)
+    expect(caseMetadata.cases[0].defendantName).toStrictEqual(defendantToInclude)
+  })
+
+  it("will match cases when the defendant name", async () => {
+    const caseMetadata = await fetchCasesAndFilter(
+      helper.postgres,
+      { defendantName: defendantToInclude.toLowerCase(), ...defaultQuery },
```
chubberlisk, line 64: This test is identical to the one above, I think we wanted to just have it as:

```suggestion
      { defendantName: defendantToInclude, ...defaultQuery },
```

packages/api/src/useCases/cases/fetchCasesAndFlitersTests/filters/defendantName.e2e.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      user
+    )
+
+    expect(caseMetadata.cases).toHaveLength(1)
+    expect(caseMetadata.totalCases).toBe(1)
+    expect(caseMetadata.returnCases).toBe(1)
+    expect(caseMetadata.cases[0].defendantName).toStrictEqual(defendantToInclude)
+  })
+
+  it("will match cases with wildcard match", async () => {
```
chubberlisk, line 74: Wondering if this more a "partial match" instead of a wildcard because it doesn't use a wildcard character (*)? 🤔

chubberlisk (commented):
Looks good to me, just two smol comments!

chubberlisk (approved):

PR #1351, opened 2025-03-12
API add get count of cases by case age

richrace:
API to return the Case Age data needed for the Case List page


richrace:
Duplication is caused by reusing the same function in API and UI. We will remove it from the UI.

richardcane (approved):

PR #1358, opened 2025-03-14
Switch to NGINX/supervisord base image

tausifpatel:


richardcane (dismissed):

packages/api/Dockerfile
```diff
@@ -54,11 +41,26 @@ FROM ${BUILD_IMAGE}
...
-COPY --from=builder /certs /certs
 
-WORKDIR /core/packages/api
+WORKDIR /app
 
 COPY --from=builder /core/node_modules/@fastify/swagger-ui/static ./static
+COPY --from=builder /core/node_modules ./node_modules
+COPY --from=builder /core/packages/common ./packages/common
+COPY --from=builder /core/packages/core ./packages/core
+COPY --from=builder /core/packages/api ./packages/api
```
richrace, line 51: We don't need to do this. On line 43 we copy the whole `core` folder

tausifpatel: Somehow it is not able to find dist folder and throws this error:
` Error: Cannot find module '/app/packages/api/dist/src/index.js'`

tausifpatel: Fixed the WORKDIR path thaat resolved the issue

richrace (approved):

PR #1366, opened 2025-03-18
Add check exceptions and triggers

richrace:
This will only return cases that have unresolved or resolved (depending on what you are filtering by) cases


packages/api/src/services/gateways/dataStoreGateways/postgres/cases/filters/exceptionsAndTriggers.ts
```diff
@@ -0,0 +1,57 @@
...
+    : ResolutionStatusNumber.Unresolved
+  let exceptionsSql
+  let triggersSql
+
+  if (userAccess(user)[Permission.Exceptions]) {
+    exceptionsSql = sql`(el.error_count > 0 AND el.error_status = ${resolutionStatus})`
+  }
+
+  if (userAccess(user)[Permission.Triggers]) {
+    const excludedTriggers = getExcludedTriggers(user.excluded_triggers?.split(","))
```
chubberlisk, line 34: Nitpick: Since `getExcludedTriggers` is just a local function, wondering whether we could move the `split` call into there so this is a bit cleaner and ends up simply being `getExcludedTriggers(user.excluded_triggers)`?

packages/api/src/useCases/cases/fetchCasesAndFiltersTests/filters/filterByReasonAndResolutionStatus.e2e.test.ts
chubberlisk, line 1: For consistency with the other file names, should we remove the `filterBy` bit?

chubberlisk (approved):

packages/api/src/useCases/cases/fetchCasesAndFiltersTests/filters/filterByReasonAndResolutionStatus.e2e.test.ts
richrace: After I do the last filter (case date range), I plan to tidy up the filters.

PR #1368, opened 2025-03-19
Fix ASN with slashes transferred from Common Platform in to the UI

Senny10:
On Common Platform, the courts will display the ASN without the "/", so it will have additional 0s to make up the characters. However, Bichard is now displaying those additional 0s plus the "/".

For example:

24/0000/00/00001234567A

It should be displayed as

24/0000/00/1234567A


packages/common/utils/getShortAsn.ts
```diff
@@ -0,0 +1,14 @@
...
+  if (!asn) {
+    return null
+  }
+
+  const asnString = asn.replace(/\//g, "")
+  if (asnString.length <= 15) {
+    return asn
+  }
+
+  return `${asnString.slice(0, 2)}/${asnString.slice(2, 6)}/${asnString.slice(6, 8)}/${asnString.slice(13)}`
```
richrace, line 11: What happens if ASN is `11/01ZD/01/400448754K`?

tausifpatel (approved):

PR #1390, opened 2025-03-28
Add ability to use the API for case list 

tausifpatel:


packages/ui/src/services/api/canUseEndpoint.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  beforeEach(() => {
+    jest.resetModules()
+  })
+
+  afterEach(() => {
+    jest.resetModules()
+  })
+
+  it("returns false when USE_API is disabled", () => {
+    mockUseApiModule(false, true, false)
```
chubberlisk, line 21:
```suggestion
  it("returns false when USE_API is disabled", () => {
    const useApi = false
    mockUseApiModule(useApi, true, false)
```

Something like this would improve readability of these tests and help remove cognitive load on the differences between each tests as it's hard to tell from the `mockUseApiModule` function call as all the arguments are booleans.

chubberlisk (approved):

packages/ui/src/services/api/canUseEndpoint.test.ts
richrace: 👍 I'll do it once it's merged 🙂

PR #1407, opened 2025-04-09
Add user comms command to CLI

joe-fol:
This PR allows us to communicate with all Bichard users using the B7 CLI tool. This cuts down our need to directly query the database for sensitive information and simplifies the process of getting this information out to the users with minimal effort. 

This tool can:
- Preview 4 predefined templates to use - Unexpected outages, Unexpected outage resolved, Planned PNC maintenance and Planned maintenance that is extended - we can add to this but I feel its a good start
- Effortlessly fetch users from DB and writes to `/tmp`  ensuring we do not have any sensitive data lingering on someone's device 
- Gives you a Preview with updated fields - user name, time of maintenance window etc 
- Ask you to type `confirm send` so there isn't an accidental mass sending of email 😅 
- Fetches our GovNotify API key `GovNotify_API_key` stored in Secrets manager in the shared AWS account
- Sends via GovNotify - we have a template folder in GovNotify called `CLI Templates` which has a template with a placeholder that we input our messages

What this tool cant do: 
- It doesn't have custom template options but we 100% could do this 
- It doesn't let you choose certain users or forces - could be useful if you needed to reach out to all supervisors or a force

This currently doesn't send real email and is only querying for Made Tech users. When we are happy we can create a Live API key in GovNotify and update in Secrets manager. 

A follow up PR will be needed to make this tool live and functional.

cli-screen-recording

I have also tested it and sent myself a Planned outage email and it looks all good.

image


chubberlisk (approved):
💯

PR #1426, opened 2025-04-28
Fix tests for new Trigger Groups

richrace:


richrace:
#1427 supersedes this PR

PR #1470, opened 2025-05-30
Fix table sort options

tausifpatel:


packages/ui/src/features/CourtCaseList/CourtCaseListTableHeader.tsx
```diff
@@ -2,25 +2,39 @@ import ColumnHeading from "features/CourtCaseFilters/ColumnHeading"
...
+
+  const ariaSort = (columnName: string) =>
+    query.orderBy === columnName ? (query.order === "asc" ? "ascending" : "descending") : "none"
+
+  const ariaLabel = (columnName: string): string => {
+    const isSorted = query.orderBy === columnName
+    const direction = ariaSort(columnName)
+    return isSorted
+      ? `${columnName} column, sorted ${direction}, click to change sort order`
+      : `${columnName} column, sortable, click to sort ascending`
```
richrace, line 37: Should this be `defendant name column, sortable, click to sort ascending` rather than `defendantName column, sortable, click to sort ascending`, for example.

tausifpatel: Sounds sensible, pushing the change. thanks :)

PR #1474, opened 2025-06-05
Add database gateway to the API package to split read and write

emadkaramad:
This PR:
- Introduces a new database gateway to separate the read and write SQL queries.
- All functions and tests in the API package have been updated to use the database gateway.
- Updated functions to return an error rather than throw an error.
- Separated types and converters for database query results, entities, and DTOs


packages/api/src/types/Case.ts
```diff
@@ -1,26 +1,61 @@
...
+  | CaseDataIndex
+
+export type CaseDataForDto = FullnameFields & Pick<Case, CaseData>
+
+export type CaseDataForIndexDto = FullnameFields &
+  Pick<Case, CaseDataIndex> & {
+    fullCount: number
+    noteCount: number
+  }
+export type CaseDataRow =
```
richrace, line 20: Whitespace

richrace (approved):
Nitpicks

PR #1476, opened 2025-06-06
Update the Architecture diagrams

richrace:
#### Hybrid Bichard

structurizr-1-HybridBichard (3)

#### Conductor

structurizr-1-Component-001


bjpirt:
A few notes:

- Should we have a user on the diagram coming in to the auth proxy?
- The ingress from Exiss actually goes:
    * Exiss -> 
    * External incoming messages s3 bucket ->
    * Message transfer lambda ->
    * External incoming messages s3 bucket ->
    * SQS event ->
    * Conductor
- Should the core worker be on the diagram? Or is it in the conductor container on the diagram?
- Now we're not running the core logic in legacy backend it doesn't actually need to connect to Beanconnect, just the PNC API
- I think Dynamo should be within the dotted line if postgres is

PR #1483, opened 2025-06-16
Use aws-sdk v3 in core

joe-fol:
This P.R updates the use of aws-sdk to v3 with the only exception of the UI. This is because we will eventually move the responsibility for creating audit logs in the UI into the API.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -322,55 +322,6 @@
         "node": ">=18.0.0"
       }
     },
-    "node_modules/@aws-sdk/client-dynamodb/node_modules/@aws-sdk/client-sso": {
-      "version": "3.828.0",
-      "resolved": "https://registry.npmjs.org/@aws-sdk/client-sso/-/client-sso-3.828.0.tgz",
-      "integrity": "sha512-qxw8JcPTaFaBwTBUr4YmLajaMh3En65SuBWAKEtjctbITRRekzR7tvr/TkwoyVOh+XoAtkwOn+BQeQbX+/wgHw==",
-      "license": "Apache-2.0",
-      "dependencies": {
-        "@aws-crypto/sha256-browser": "5.2.0",
-        "@aws-crypto/sha256-js": "5.2.0",
-        "@aws-sdk/core": "3.826.0",
-        "@aws-sdk/middleware-host-header": "3.821.0",
-        "@aws-sdk/middleware-logger": "3.821.0",
-        "@aws-sdk/middleware-recursion-detection": "3.821.0",
-        "@aws-sdk/middleware-user-agent": "3.828.0",
-        "@aws-sdk/region-config-resolver": "3.821.0",
-        "@aws-sdk/types": "3.821.0",
-        "@aws-sdk/util-endpoints": "3.828.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -70,6 +70,10 @@
     "url": "https://github.com/ministryofjustice/bichard7-next-core/issues"
   },
   "devDependencies": {
+    "@aws-sdk/client-dynamodb": "^3.821.0",
+    "@aws-sdk/client-lambda": "^3.821.0",
+    "@aws-sdk/client-rds": "^3.821.0",
+    "@aws-sdk/lib-dynamodb": "^3.821.0",
```
richrace, line 76: Why have you put these here?

packages/api/src/services/gateways/dynamo/AuditLogDynamoGateway/AuditLogDynamoGateway.ts
```diff
@@ -470,6 +470,10 @@ export default class AuditLogDynamoGateway extends DynamoGateway implements Audi
       return dynamoUpdates
     }
 
+    if (!dynamoUpdates) {
+      return dynamoUpdates
+    }
```
richrace, line 475: Do you know if this needs to be here? We create `dynamoUpdates` as an array, and the following line handles an empty array.

packages/api/src/tests/testGateways/TestDynamoGateway/getTableAttributes.ts
```diff
@@ -1,20 +1,21 @@
...
+  const attributesInIndexes: AttributeDefinition[] = []
 
   secondaryIndexes.forEach((index) => {
     const { hashKey, rangeKey } = index
+
     if (
       hashKey !== partitionKey &&
       hashKey !== sortKey &&
-      attributesInIndexes.filter((x) => x.AttributeName === hashKey).length === 0
+      !attributesInIndexes.some((x) => x.AttributeName === hashKey)
```
richrace, line 18: Suggestion: I personally do not negate the `attributesInIndex`

packages/api/src/tests/testGateways/TestDynamoGateway/getTableAttributes.ts
```diff
@@ -23,10 +24,10 @@ const getTableAttributes = (
     }
 
     if (
-      !!rangeKey &&
+      rangeKey &&
       rangeKey !== partitionKey &&
       rangeKey !== sortKey &&
-      attributesInIndexes.filter((x) => x.AttributeName === rangeKey).length === 0
+      !attributesInIndexes.some((x) => x.AttributeName === rangeKey)
```
richrace, line 30: Same here

packages/api/src/services/gateways/dynamo/AuditLogDynamoGateway/AuditLogDynamoGateway.ts
joe-fol: The return type has changed, and it can now return undefined

richrace (approved):

richrace (approved):

PR #1495, opened 2025-06-26
Remove padding from the page layout on smaller screen size

tausifpatel:


packages/ui/src/components/Header/Header.styles.tsx
```diff
@@ -3,6 +3,10 @@ import styled from "styled-components"
 const MojHeaderContainer = styled.div`
   max-width: 100%;
   padding: 0 40px;
+
+  @media (max-width: 1019px) {
```
emadkaramad, line 7: It would be nice and make the code more readable if we used constant variables for the screen sizes.

emadkaramad (approved):

PR #1508, opened 2025-07-22
Fix listener

richrace:
In the Refresh Button component, we didn't stop listening to an event. This fixes that.


packages/ui/src/components/Buttons/RefreshButton.tsx
```diff
@@ -11,33 +11,42 @@ interface RefreshButtonProps extends React.ComponentProps<"button"> {
...
+  useEffect(() => {
     const interval = setInterval(() => {
       setTimeAgo(formatDistanceStrict(dateAgo, new Date()))
     }, 1000)
 
     return () => clearInterval(interval)
-  }, [dateAgo, timeAgo, nextRouterChanged])
+  }, [dateAgo])
+
+  useEffect(() => {
```
tomwatkins1994, line 35: Could the logic in this `useEffect` just be merged into the `handleRouteChange` function on line 18 perhaps? We could then just drop this effect as well as the `nextRouterChanged` state.

richrace: If I remember correctly, whilst looking into this #1505, it needed to be in a separate `useEffect` to work.

PR #1525, opened 2025-08-01
Use new Audit Log API for 

richrace:
This PR uses the Bichard API to create and update Audit Logs.


packages/api/src/routes/v1/audit-logs/createAuditLog.ts
```diff
@@ -32,7 +32,7 @@ const schema = {
   ...auth,
   body: InputApiAuditLogSchema,
   response: {
-    [CREATED]: z.null().meta({ description: "No content" }),
+    [CREATED]: OutputApiAuditLogSchema.meta({ description: "No content" }),
```
tomwatkins1994, line 35: Should we update the description in the meta tag?

packages/common/AuditLogApiClient/createApiConfig.ts
```diff
@@ -6,8 +10,24 @@ export type AuditLogApiConfig = {
 
 const createApiConfig = (): AuditLogApiConfig => {
   const apiUrl = process.env.AUDIT_LOG_API_URL
```
tomwatkins1994, line 12: Could we do the check on `apiUrl` here rather than potentially doing it after generating a JWT which wouldn't then get used if `apiUrl` is blank?

packages/common/AuditLogApiClient/createApiConfig.ts
```diff
@@ -6,8 +10,24 @@ export type AuditLogApiConfig = {
...
-  const basePath = process.env.AUDIT_LOG_API_BASE_PATH ?? "messages"
-  const apiKey = process.env.AUDIT_LOG_API_KEY
+  let basePath = "messages"
+  let apiKey = process.env.AUDIT_LOG_API_KEY
```
tomwatkins1994, line 14: Could we set `apiKey` to the environment variable in the else block instead? Just thinking that we could have a misconfig that doesn't result in an error being thrown because `AUDIT_LOG_API_URL` env variable could be set to the new API URL and `AUDIT_LOG_API_KEY` could be set to the old API key, `AUDIT_LOG_USE_JWT` could also be set as true but perhaps `AUTH_JWT_SECRET` has been accidentally set as blank. In this case it would not error and would just be incorrectly sending the wrong API key to the API instead of the JWT.

packages/api/src/routes/v1/audit-logs/createAuditLog.ts
richrace: Fixed

tomwatkins1994 (approved):

PR #1530, opened 2025-08-04
Make the key information box collapsible when the page is viewed at 150% zoom

richrace:
### Hide

Screenshot 2025-08-08 at 10 53 22

### Show

Screenshot 2025-08-08 at 10 53 32

### DIFF
--- a/packages/ui/src/features/CourtCaseDetails/Header.styles.tsx
+++ b/packages/ui/src/features/CourtCaseDetails/Header.styles.tsx
@@ -35,7 +35,7 @@ const CaseDetailHeaderRow = styled.div`
     cursor: pointer;
     margin-bottom: 10px;
 
-    @media (min-resolution: 144dpi) and (max-width: 768px) {
+    @media (min-resolution: 144dpi) and (max-width: ${breakpoints.compact}) {
       visibility: visible;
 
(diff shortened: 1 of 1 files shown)


packages/ui/src/features/CourtCaseDetails/Header.tsx
```diff
@@ -73,6 +78,24 @@ const Header: React.FC<Props> = ({ canReallocate }: Props) => {
...
+  useEffect(() => {
+    if (typeof window !== "undefined") {
+      const handleResize = () => {
+        setWindowWidth(window.innerWidth)
+      }
+
+      window.addEventListener("resize", handleResize)
+
+      if (windowWidth > 768) {
+        setIsContentVisible(true)
```
tomwatkins1994, line 90: Won't this present the _potential_ issue that if someone hides the box but then resizes the box will then show again?

richrace: It's needed because if I hide it and then resize my browser to, say, 1024px, I will lose the button to show it.

tomwatkins1994: Gotcha, makes sense to me. Everything else LGTM

PR #1564, opened 2025-08-21
BICAWS7-3629 Enable API by Force

emmalanigan:


packages/ui/src/services/api/canUseEndpoint.ts
```diff
@@ -1,12 +1,12 @@
...
 
 export enum ApiEndpoints {
   CaseDetails = "CaseDetails",
   CaseList = "CaseList"
 }
 
-export const canUseApiEndpoint = (endpoint: ApiEndpoints): boolean => {
-  if (!USE_API) {
+export const canUseApiEndpoint = (endpoint: ApiEndpoints, forces: string[]): boolean => {
+  if (!USE_API || !forces.some((force) => FORCES_WITH_API_ENABLED.has(force))) {
```
richrace, line 9: Can we split these checks out? 

We want `USE_API` to be a global enable for everyone, regardless of whether the force can use the API.

PR #1576, opened 2025-09-01
BICAWS7-3638 - Add user level and force level feature flags for court date / message received date mismatch filtering

tomwatkins1994:
* Add force level feature flag for court date / message received date mismatch filtering
* Add function to check the user level feature flag for court date / message received date mismatch filtering and if their visible forces include one that is enabled in force level flag
* Flag exists in new `features/flags` directory


packages/ui/src/services/entities/User.ts
```diff
@@ -40,7 +40,7 @@ export default class User {
   excludedTriggers!: string[]
 
   @Column({ name: "feature_flags", transformer: jsonTransformer, type: "jsonb" })
-  featureFlags!: Record<string, boolean>
+  featureFlags!: { useCourtDateReceivedDateMismatchFilter?: boolean } & Record<string, boolean>
```
richrace, line 43: We don't tend to hardcode user feature flags.

richrace (approved):

PR #1593, opened 2025-09-15
BICAWS7-3686 Add user level and force level feature flags for trigger and exception quality auditing

emmalanigan:
- Add force level feature flag for trigger and exception quality auditing
- Add function to check the user level feature flag for trigger and exception quality auditing and if their visible forces include one that is enabled in force level flag


.circleci/continue_config.yml
tomwatkins1994, line 1: Think uou're missing the definition for the `FORCES_WITH_TRIGGER_AND_EXCEPTION_QUALITY_AUDITING_ENABLED` param in the `start_bichard7_legacy` task

tomwatkins1994 (changes requested):

tomwatkins1994 (approved):

PR #1602, opened 2025-09-18
Add FormGroup component

tomwatkins1994:
Adds a reusable form group component:

* Ensures we always use the right GDS class
* Adds error class functionality

### DIFF
--- a/packages/ui/src/components/NoteTextArea.tsx
+++ b/packages/ui/src/components/NoteTextArea.tsx
@@ -30,7 +30,7 @@ export const NoteTextArea = ({
   maxLength = MAX_NOTE_LENGTH
 }: Props) => {
   return (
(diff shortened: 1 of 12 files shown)


packages/ui/src/components/NoteTextArea.tsx
```diff
@@ -29,7 +30,7 @@ export const NoteTextArea = ({
   maxLength = MAX_NOTE_LENGTH
 }: Props) => {
   return (
-    <div id={id} className={"govuk-form-group" + (showError ? " govuk-form-group--error" : "")}>
+    <FormGroup id={id} showError>
```
richrace, line 33: What happens if `showError` is false? It's in your test.

richrace (approved):

PR #1605, opened 2025-09-19
BICAWS7-3640 UI Add the checkbox and wording from the design to case list filters

emmalanigan:


packages/ui/src/config.ts
```diff
@@ -20,7 +20,7 @@ export const FORCES_WITH_API_ENABLED: Set<string> = new Set(
   (process.env.FORCES_WITH_API_ENABLED ?? "").split(",").filter(Boolean)
 )
 export const FORCES_WITH_COURT_DATE_RECEIVED_DATE_MISMATCH_ENABLED: Set<string> = new Set(
-  (process.env.FORCES_WITH_COURT_DATE_RECEIVED_DATE_MISMATCH_ENABLED ?? "").split(",").filter(Boolean)
+  (process.env.FORCES_WITH_COURT_DATE_RECEIVED_DATE_MISMATCH_ENABLED ?? "01").split(",").filter(Boolean)
```
richrace, line 23: I don't think this should be enabled by default

richrace (changes requested):

packages/ui/src/components/SearchFilters/CourtDateFilter.tsx
```diff
@@ -104,6 +113,13 @@ const CourtDateFilter: React.FC<Props> = ({ caseAges, caseAgeCounts, dispatch, d
             </ScrollableCaseAgesContainer>
           </div>
         </div>
+        <ConditionalRender isRendered={canUseCourtDateReceivedDateMismatchFilters}>
+          <CourtDateReceivedDateMismatchCheckbox
+            id={"case-show-date-received-mismatch"}
+            // value={dateRange?.to}
```
richrace, line 119: Comment

PR #1608, opened 2025-09-19
BICAWS7-3754 - Add auditing box

tomwatkins1994:
This PR adds all the components needed for audit but does not add them to the case details screen. Component tests mocking the API calls currently, I will add the actual API route in a future PR.

### DIFF
--- a/packages/ui/src/features/CourtCaseDetails/Sidebar/Audit/ExceptionQualityDropdown.tsx
+++ b/packages/ui/src/features/CourtCaseDetails/Sidebar/Audit/ExceptionQualityDropdown.tsx
@@ -5,7 +5,7 @@ import { FormGroup } from "components/FormGroup"
 
 export type ExceptionQualityDropdownProps = ComponentProps<typeof Select>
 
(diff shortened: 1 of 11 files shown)


packages/ui/src/features/CourtCaseDetails/Sidebar/Audit/ExceptionQualityDropdown.tsx
```diff
@@ -0,0 +1,36 @@
...
+    <FormGroup>
+      <label className={"govuk-visually-hidden"} htmlFor={"exception-quality"}>
+        {"Set Exception Quality"}
+      </label>
+      <Select
+        {...props}
+        placeholder={"Set Exception Quality"}
+        name={"exception-quality"}
+        aria-invalid={props.showError}
+        aria-describedby="exception-quality-error"
```
tausifpatel, line 21: Should `aria-describedby="exception-quality-error"` only be rendered when there’s an error, instead of always being present?

tausifpatel (approved):

PR #1617, opened 2025-09-26
Save audit note

tausifpatel:
[User story](https://dsdmoj.atlassian.net/jira/software/c/projects/BICAWS7/boards/1404?selectedIssue=BICAWS7-3753)

This PR:
- Adds a new service for creating notes.
- Updates the `saveAuditResults` use case to handle updating exceptionQuality, triggerQuality, and creating notes within a single transaction, ensuring that if one of them fails, all changes are rolled back.
- Updates the POST endpoint payload to accept a note.


packages/api/src/useCases/cases/saveAuditResults.ts
```diff
@@ -1,29 +1,63 @@
 import type { PromiseResult } from "@moj-bichard7/common/types/Result"
 
+import { exceptionQualityValues } from "@moj-bichard7/common/types/ExceptionQuality"
 import { isError } from "@moj-bichard7/common/types/Result"
+import { triggerQualityValues } from "@moj-bichard7/common/types/TriggerQuality"
 
 import type { AuditQuality } from "../../services/db/cases/auditCase"
```
tomwatkins1994, line 7: Could we apply the types to `AuditQuality` so we don't have to type cast things?

packages/api/src/useCases/cases/saveAuditResults.ts
```diff
@@ -1,29 +1,63 @@
...
 import type { WritableDatabaseConnection } from "../../types/DatabaseGateway"
 
 import auditCase from "../../services/db/cases/auditCase"
+import insertNote from "../../services/db/cases/insertNote"
+import { NotFoundError } from "../../types/errors/NotFoundError"
 import { UnprocessableEntityError } from "../../types/errors/UnprocessableEntityError"
 
+const formatNote = (triggerQuality?: number, errorQuality?: number, note?: string): string => {
```
tomwatkins1994, line 15: Could the params be of type `TriggerQuality` and `ExceptionQuality`? Then we can avoid the type cast

packages/api/src/useCases/cases/saveAuditResults.ts
tausifpatel: Pushed the changes

tomwatkins1994 (approved):

PR #1622, opened 2025-09-29
BICAWS7-3754 UI add auditing box to case details screen

emmalanigan:
Add auditing box component to the case details screen
Add logic around when this will be visible:

- Feature flags must be enabled
- Force must be allowed
- User should be in the correct security group


packages/ui/cypress/fixtures/users.ts
```diff
@@ -49,6 +49,32 @@ const users: Record<string, Partial<User> & { groups: UserGroup[] }> = {
...
+    surname: "Supervisor",
+    email: "auditSupervisor@example.com",
+    password: hashedPassword,
+    groups: [UserGroup.Audit, UserGroup.AuditLoggingManager],
+    featureFlags: {
+      useCourtDateReceivedDateMismatchFiltersEnabled: true,
+      useTriggerAndExceptionQualityAuditingEnabled: true
+    }
+  },
+  UserWithoutAuditGroups: {
```
tomwatkins1994, line 65: We could probably just reuse the general handler user here as that isn't in the audit groups

packages/ui/cypress/fixtures/users.ts
```diff
@@ -49,6 +49,32 @@ const users: Record<string, Partial<User> & { groups: UserGroup[] }> = {
...
     }
   },
+  AuditSupervisor: {
+    username: "AuditSupervisor",
+    visibleForces: ["001"],
+    forenames: "Audit",
+    surname: "Supervisor",
+    email: "auditSupervisor@example.com",
+    password: hashedPassword,
+    groups: [UserGroup.Audit, UserGroup.AuditLoggingManager],
```
tomwatkins1994, line 59: I think we will need `UserGroup.NewUI` here?

packages/ui/src/pages/court-cases/[courtCaseId]/index.tsx
```diff
@@ -44,6 +44,7 @@ import logger from "utils/logger"
 import { logRenderTime } from "utils/logging"
 import redirectTo from "utils/redirectTo"
 import shouldShowSwitchingFeedbackForm from "utils/shouldShowSwitchingFeedbackForm"
+import { canUseTriggerAndExceptionQualityAuditing } from "../../../features/flags/canUseTriggerAndExceptionQualityAuditing"
```
Comment (tomwatkins1994) [inline packages/ui/src/pages/court-cases/[courtCaseId]/index.tsx:47]: We can just import this from `features/flags/canUseTriggerAndExceptionQualityAuditing` without needing to go back directories

tomwatkins1994 (approved):

PR #1628, opened 2025-10-02
BICAWS7-3757  Only show audit box if user is in correct security group

emmalanigan:
After originally setting security groups to Audit and AuditLoggingManager, we established it would actually make more sense for the security group to be Supervisor in line with the current set up of Old Bichard.

This PR updates tests and logic associated with checking the security groups.


packages/ui/cypress/e2e/case-details/view-case-details/sidebar.cy.ts
```diff
@@ -69,7 +69,7 @@ describe("sidebar", () => {
 
   describe("quality status card", () => {
     it("Should show qualityStatusCard when feature flags enabled and user has audit groups", () => {
```
tomwatkins1994, line 71: *when user is a supervisor

tomwatkins1994 (changes requested):

tomwatkins1994 (approved):

PR #1635, opened 2025-10-08
Handle invalid UUID

richrace:
Relax the Zod parsing as we have some `message_id` that match UUIDs


tomwatkins1994:
Why not just change it to be `z.string()` since `z.uuid()` is just a narrower string type?

PR #1911, opened 2026-04-10
Remove axios npm package for UI, User Service, API and Conductor

richrace:
NPM's axios keeps getting CVEs, so according to [ADR 005](https://github.com/ministryofjustice/bichard7-next-core/blob/main/docs/architecture/adrs/0005-using-external-libraries.md), we should use `fetch`, as it has no dependencies and is part of Node. Or use `undici` if needed.

- [x] User Service
- [x] UI
- [x] Conductor 
- [x] API

---

These packages are quite large, and they will be in separate PRs

- Core
- Common
- E2E Tests


packages/ui/src/components/Typeaheads/ForceOwnerTypeahead.tsx
```diff
@@ -15,19 +14,31 @@ const ForceOwnerTypeahead: React.FC<Props> = ({ onSelect, currentForceOwner }: P
...
+
+        if (currentForceOwner) {
+          params.append("currentForceOwner", currentForceOwner)
+        }
+
+        if (searchStringParam) {
+          params.append("search", searchStringParam)
+        }
+
+        const response = await fetch(`/bichard/api/force-owner?${params}`, {
```
tomwatkins1994: We should probably use `params.toString()` here to handle encoding. Also if the params are empty it would leave a `?` on the end with no params, best to add that only if there are param values.

tomwatkins1994: Also apply this logic to other files doing the same thing

packages/ui/test/helpers/fetchAuditLogEvents.ts
tomwatkins1994, line 1: The axios version here would throw on an error, the fetch would just fail silenty. We should get for an error and throw it to keep the behaviour consistent

tomwatkins1994 (changes requested):

packages/ui/test/helpers/fetchAuditLogEvents.ts
richrace: Will look into it.

tomwatkins1994 (approved):

PR #1924, opened 2026-04-16
Tweak the retry transaction function

richrace:
We saw 1 transaction fail. This PR addresses that.

* Add another retry
* Increase the timeout 
* Refactor `logger` so it accurately reflects the outcome

### DIFF
--- a/packages/ui/src/services/retryTransaction.ts
+++ b/packages/ui/src/services/retryTransaction.ts
@@ -19,25 +19,27 @@ export const retryTransaction = async <T, Args extends unknown[]>(
   callback: (...args: Args) => Promise<T>,
   ...callbackArgs: Args
 ): Promise<T> => {
-  const maxRetries = 2
+  const maxRetries = 3
   let attempt = 0
-  let lastError: unknown
 
   while (attempt < maxRetries) {
     try {
       return await callback(...callbackArgs)
     } catch (error) {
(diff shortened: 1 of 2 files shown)


packages/ui/src/services/retryTransaction.ts
```diff
@@ -19,25 +19,27 @@ export const retryTransaction = async <T, Args extends unknown[]>(
...
 
       if (!isRetryableError(error) || attempt >= maxRetries - 1) {
+        logger.error(`Error inside transaction (retry: ${attempt}): ${errorForLog.name} - ${errorForLog.message}`)
         throw error
       }
 
-      await delay(randomInt(0, 100) * (attempt + 1))
+      logger.warn(`Error inside transaction (retry: ${attempt}): ${errorForLog.name} - ${errorForLog.message}`)
+
+      await delay(randomInt(50, 150) * (attempt + 1))
```
timhowes-lab: Out of interest, why are we using a random int rather than a fixed?

richrace: We are creating jitter, we want the number to be random so that it doesn't happen at the same time. E.g. we have multiple DB transactions, we need to add a bit of randomness so they don't fire at the same time.

And we can't use `Math.random()` (security reasons).

timhowes-lab (approved):

PR #1933, opened 2026-04-20
Update postgres package

richrace:
* Refactor the transaction mechanics as the type was updated
* Fix open handles - the test DB was not closed


timhowes-lab (dismissed):

packages/api/src/types/DatabaseGateway.ts
tomwatkins1994, line 1: Something about `TransactionDatabaseConnection` doesn't feel quite right to me. For a few reasons:

* Transactions aren't only used for writes, there are legit use cases for using them for read only stuff like querying multiple tables and getting consistent results
* It obscures that it is a writeable connection 
* The fact it is also then mapped to the `writeable` property of the gateway instead of the `WritableDatabaseConnection` feels a bit odd
* Is the thought that we just don't want to be able to have nested transactions?

richrace: Yes, it's a bit clunky. 

The problem is that the transaction returns a different type.

Thanks for the feedback, I will see if I can make a better/more visible solution.

richrace: To address points:

* The `WritableDatabaseConnection` is using the Writer RDS instance.
* We can use the `database.readonly` to create transactions `database.readonly.connection.begin()`
* We don't have transaction options on the `postgres.TractionSql` type (`postgres` npm decision)
* If we want to use the helper function of the `WritableDatabaseConnection`, we can. Or if we want to use your own, you can 
  * `database.writable.connection.begin()`
  * `database.writable.transaction()`

I've made some changes to hopefully make it more readable.

tomwatkins1994 (approved):

PR #1940, opened 2026-04-24
Bicaws7 4154 remove axios from common

timhowes-lab:


packages/common/utils/dateTransformer.ts
tomwatkins1994, line 1: Are there any tests for this code so we can feel safe with this refactor?

packages/common/package.json
```diff
@@ -52,6 +51,7 @@
     "postgres": "3.4.9",
     "promise-poller": "^1.9.1",
     "stompit": "^1.0.0",
+    "undici": "^7.25.0",
```
tomwatkins1994: We have pinned `undici` to v6 elsewhere as v7 drops support for node 20, we should change this to v6 for the same reason (for now)

packages/common/utils/dateTransformer.ts
tomwatkins1994: Also `date-fns` has a built in `isValid` function so I wonder if we could just that instead?

tomwatkins1994 (changes requested):

packages/common/package.json
```diff
@@ -52,6 +51,7 @@
     "postgres": "3.4.9",
     "promise-poller": "^1.9.1",
     "stompit": "^1.0.0",
+    "undici": "^6.25.0",
```
richrace, line 54: You need to pin this in `.ncurc` to stop it updating

.ncurc.js
tomwatkins1994, line 1: I think we may need this in a `packages/common/.ncurc`? (I know this doesn't exist currently. @richrace do you know if putting it globally like this works?

timhowes-lab: "update-deps": "ncu -u --deep && npm install",

Am I right in thinking the --deep should mean it affects nested package.jsons?

richrace: If an `.ncurc` file exists in the root, it doesn't need the rc file in the monorepo package.

If there is an rc file in the monorepo package, it overrides the root one.

tomwatkins1994 (approved):

PR #1952, opened 2026-04-30
Fix flaky e2e tests

tausifpatel:
Some of the e2e tests were failing on CircleCI, and it became a nightmare whenever a PR was removed from the merge queue, we repeatedly had to add it back. We suspected that Conductor was not clearing workflows properly, which appeared to cause issues for e2e tests that receive more than one message (for example test 143).

To address this, Conductor workflows will now be cleared before each test run. This should help reduce the flakiness of e2e test runs on CircleCI.


packages/e2e-test/features/027-orphaned-non-recordable/test.feature
```diff
@@ -23,9 +23,10 @@ Feature: {027} R3.2_Orphaned Non-Recordable Offence
 	Scenario: No exceptions are created for orphaned non-recordable offence
 		When "input-message-1" is received
 			And I am logged in as "supervisor"
+		Then I wait "4" seconds
```
richrace: Is this meant to be here?

packages/e2e-test/utils/conductor.ts
```diff
@@ -19,7 +20,7 @@ const conductorApi = axios.create({
 
 const fetchCompletedBichardProcessWorkflow = async (workflowId: string) => {
   const workflow = await conductorApi
-    .get(`${CONDUCTOR_API_URL}/api/workflow/${workflowId}`)
+    .get(`${CONDUCTOR_URL}/api/workflow/${workflowId}`)
```
richrace: This change in `packages/e2e-test/helpers/ConductorHelper.ts` is `CONDUCTOR_URL: process.env.CONDUCTOR_URL ?? "http://localhost:5002/api",`

So `/api/api/`?

packages/e2e-test/features/027-orphaned-non-recordable/test.feature
tausifpatel: Yes it has been kept there deliberately to ensure conductor workflow finishes.

packages/e2e-test/utils/conductor.ts
tausifpatel: Good spot, I have pushed the change.

packages/e2e-test/features/027-orphaned-non-recordable/test.feature
richrace: Can't we do something a bit more clever? i.e., can we wait for conductor workflows to finish?

packages/e2e-test/utils/conductor.ts
richrace: There are many URLs in this file that need updating.

packages/e2e-test/features/027-orphaned-non-recordable/test.feature
tausifpatel: Thats a good shout, new changes are pushed reflecting your suggestion.

richrace (dismissed):

emadkaramad (approved):

PR #1957, opened 2026-05-05
Fix LEDS seeding

emadkaramad:


packages/core/lib/policeGateway/leds/preProcessPersonUrn.test.ts
```diff
@@ -2,7 +2,7 @@ import preProcessPersonUrn from "./preProcessPersonUrn"
 
 describe("preProcessPersonUrn", () => {
   it("should use 2-digit year format and remove leading zeros from id", () => {
```
tausifpatel: Can we change title to _should use 4-digit year format_

emadkaramad: Good shout! Fixed.

emadkaramad: Changed it to "the same year format" as the logic doesn't convert the year format.

tausifpatel (dismissed):

tausifpatel (approved):

PR #1965, opened 2026-05-06
Bicaws7 4178 UI automated reports- added automation rate and top exceptions

timhowes-lab:
-Added new 'Automated report type' string union type and map. Adding the new types to the existing 'Report type' string union and map would have required changes to the API code
-Added two new report types to the reports page dropdown
-Date range filters are now hidden when 'Clear filters' is clicked. Previously these fields persisted, but now that we have report types that don't have a date range, makes more sense to hide them
-'Run report' button is now only visible when a non-automated report type has been selected
-Updated audit logging to allow for XLSX files


packages/ui/cypress/e2e/reports/automation-rate.cy.ts
```diff
@@ -0,0 +1,16 @@
+import { loginAndVisit } from "../../support/helpers"
```
richrace, line 1: This doesn't need the B7 API to run `automation-rate.api.cy.ts`; the `api` in the file name is only run on the Cypress tests that need the API.

packages/ui/cypress/e2e/reports/automation-rate.cy.ts
```diff
@@ -0,0 +1,16 @@
...
+    loginAndVisit("Supervisor", "/bichard/report-selection")
+  })
+
+  it("should display only 'Download report' button when 'Automation Rate' is selected from dropdown", () => {
+    cy.get("#report-select").select("Automation Rate")
+    cy.get("#date-from").should("not.exist")
+    cy.get("#date-to").should("not.exist")
+    cy.get("#exceptions").should("not.exist")
+    cy.get("#triggers").should("not.exist")
+    cy.get("#download-automated-report").should("exist")
```
richrace, line 14: We should add a test that is pointing the right URL

packages/ui/cypress/e2e/reports/top-exceptions.cy.ts
```diff
@@ -0,0 +1,16 @@
+import { loginAndVisit } from "../../support/helpers"
```
richrace, line 1: Same comments as above

richrace (changes requested):

packages/ui/cypress/component/reports/FilterSelectReportSection.cy.tsx
```diff
@@ -5,17 +5,8 @@ describe("ReportSelectionFilter", () => {
     cy.intercept("GET", `**/bichard/api/reports*`).as("downloadApi")
   })
 
-  const apiCallCheck = (shouldRun: boolean) => {
```
richrace, line 8: We no longer have a test for entering date values and clicking the button

packages/ui/cypress/component/reports/FilterSelectReportSection.cy.tsx
```diff
@@ -24,31 +15,39 @@ describe("ReportSelectionFilter", () => {
...
   })
 
-  it("'This field is required' message is displayed when no report is selected and 'Run report' is clicked", () => {
+  it("'Run report' button is hidden when no report is selected", () => {
     cy.mount(<ReportSelectionFilter />)
+    cy.get("button#run-report").should("not.exist")
+  })
 
-    cy.get("button#run-report").click()
-    cy.get("div#report-section").find("p.govuk-error-message").should("contain", "This field is required")
```
richrace, line 37: We no longer have a test for the error messages

richrace (commented):
Just that one file, we've lost a lot of tests.

We need to handle the audit logging. It just handles the CSV audit.

packages/ui/cypress/component/reports/FilterSelectReportSection.cy.tsx
timhowes-lab: The 'Run report' button is now only visible when a report type has been selected (and specifically only a non-automated report type), whereas before it was always visible, even when no report type had been selected. This means it's not possible to click the 'Run report' button without first selecting a report type, so the error message will never be shown. So I've changed this test to instead test the visibility of the button rather than check for an error message

packages/ui/cypress/component/reports/FilterSelectReportSection.cy.tsx
timhowes-lab: We have a separate test file FilterDateRangeSection.cy.tsx for this. The logic I've removed here wasn't testing the date fields directly, it was just to populate the date fields to allow testing the Report type input field in isolation

packages/ui/src/services/reports/auditLogFileDownload.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  const { csvDownload, xlsxDownload, reportType, fromDate, toDate } = query
+
+  if (csvDownload !== "true" && xlsxDownload !== "true") {
+    return new Error("400")
+  }
+
+  if (csvDownload === "true" && xlsxDownload === "true") {
+    return new Error("400")
+  }
```
richrace, line 39: Is one meant to be `false`?

timhowes-lab: So in the first case if both are not true, then we don't know what file is being specified, so error. 

And in the second case, if both are true then two different files are being specified at the same time, so also error

richrace (approved):

PR #1971, opened 2026-05-07
Add User Performance Detail report to API

richrace:
* Add Map on boot for Exceptions and Triggers with their descriptions from data package (allows for `O(1)` access)
* Add the Report

### DIFF
--- a/packages/api/src/useCases/data/getExceptionDescriptionMap.ts
+++ b/packages/api/src/useCases/data/getExceptionDescriptionMap.ts
@@ -1,0 +1,9 @@
+import exceptionDefinitions from "@moj-bichard7-developers/bichard7-next-data/dist/data/exception-definitions.json"
+
+const exceptionMap = new Map<string, string>()
+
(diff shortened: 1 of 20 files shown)


packages/api/src/useCases/data/getExceptionDescriptionMap.ts
tomwatkins1994, line 1: Maps can be created inline like:

```ts
const exceptionMap = new Map<string, string>(
   Object.values(exceptionDefinitions).map(def => ([def.code, def.shortDescription ?? def.description ?? "Unknown Exception"]))
)
```

Might be better?

packages/api/src/useCases/data/getTriggerDescriptionMap.ts
tomwatkins1994, line 1: See comment on `packages/api/src/useCases/data/getExceptionDescriptionMap.ts`

tomwatkins1994 (approved):

PR #1982, opened 2026-05-13
Allow reports date range to back 12 months

richrace:
- Allows users to go back to the start of the month 12 months ago
- Auto fills the `dateTo` field with the end of the month or today's date
- Limits the `dateTo` to a calendar month


packages/ui/src/utils/reports/dateRange.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  jest.useRealTimers()
+})
+
+describe("dateRange", () => {
+  describe("when dateFrom is beyond the 12-month limit", () => {
+    it("returns an error", () => {
+      const result = dateRange(parseISO("2025-04-30"))
+
+      expect(result).toBeInstanceOf(Error)
+      expect((result as Error).message).toBe("Date cannot be more than 12 months")
```
smlbrd:
```suggestion
      expect((result as Error).message).toBe("Date cannot be more than 12 months ago")
```

packages/ui/cypress/component/reports/FilterDateRangeSection.cy.tsx
```diff
@@ -1,139 +1,205 @@
...
   })
 
-  it("'This field is required' message is displayed when 'Date from' has no value and 'Run report' is clicked", () => {
-    cy.mount(<ReportSelectionFilter />)
-    cy.get('select[name="select-case-type"]').select("Warrants")
-    cy.get("button#run-report").click()
-    cy.get("div#report-selection-date-from").find("p.govuk-error-message").should("contain", "This field is required")
-    apiCallCheck(false)
+  describe("valid submission", () => {
+    it("calls the API when a valid date range is entered and 'Run report' is clicked", () => {
```
tomwatkins1994: Should valid submission test the earliest date over todays date?

packages/ui/src/utils/reports/dateRange.test.ts
smlbrd: I had to read this a couple of times - some indicator of time span might make it a bit clearer?

"Please select a date within the last 12 months" might also be an option

packages/ui/src/utils/reports/validateDateRange.ts
```diff
@@ -13,12 +18,22 @@ export const validateDateRange = (dateFromStr: string, dateToStr: string) => {
...
+  if (isAfter(dateFrom, dateTo)) {
     return {
       fromError: DATE_CANNOT_BE_AFTER_DATE_TO,
       toError: DATE_CANNOT_BE_BEFORE_DATE_FROM
     }
   }
 
+  const range = dateRange(dateFrom)
+
+  if (range instanceof Error) {
```
tomwatkins1994: Should we use the `isError` function here?

packages/ui/src/utils/reports/dateRange.ts
```diff
@@ -0,0 +1,24 @@
...
+  startDate: Date
+  endDate: Date
+}
+
+export const dateRange = (dateFrom: Date): Result<DateRange> => {
+  const today = new Date()
+  const maxFromDate = startOfMonth(subMonths(today, 12))
+
+  if (isBefore(dateFrom, maxFromDate)) {
+    return new Error("Date cannot be more than 12 months")
```
smlbrd: As above!

richrace: This is an internal error that (should) never be seen by the user. It is used as the [Result type pattern](https://en.wikipedia.org/wiki/Result_type).

https://github.com/ministryofjustice/bichard7-next-core/pull/1982/changes#diff-44c0847945bc3f40baa6a0d8760ff16743bb0e1d05d4e3e274ee7cc3e740c50bR28-R32

I've made the change to add ` ago`

packages/ui/cypress/component/reports/FilterDateRangeSection.cy.tsx
richrace: I will add another test.

packages/ui/src/utils/reports/dateRange.ts
richrace: These are the error the user sees: https://github.com/ministryofjustice/bichard7-next-core/blob/ce7bbe5d4c9183f736a634c9973d876a12f3ecc7/packages/ui/src/utils/reports/validationMessages.ts#L9-L8

tomwatkins1994 (approved):

PR #1983, opened 2026-05-14
Bicaws7 4177 UI add report user performance detail

timhowes-lab:


packages/ui/cypress/e2e/reports/user-performance-detail.api.cy.ts
```diff
@@ -0,0 +1,91 @@
...
+        })
+
+        for (let i = 1; i < 7; i++) {
+          const pastDate = format(subDays(today, i), "dd/MM/yyyy")
+
+          cy.contains("h3", pastDate)
+            .parent("section")
+            .within(() => {
+              cy.get("table").should("not.exist")
+              cy.get(".ReportTable-styles__StyledReportTable-sc-65437099-0").should("not.exist")
```
katerd: Is there another selector which could be used? It seems a little brittle? It might not be, but the `sc-65427...` part looks curious

timhowes-lab: Yeh good shout I'll see if it can be replaced with something more generic

katerd (approved):

PR #1984, opened 2026-05-15
Bicaws7 4261 UI collapsible cells for offence titles wordings and notes

timhowes-lab:


packages/ui/cypress/component/reports/ExpandableCell.cy.tsx
tomwatkins1994, line 1: The file is called `ExpandableCell.cy.tsx` but seems to be testing `ReportTableRow`?

packages/ui/src/components/Reports/ExpandableCell.tsx
tomwatkins1994, line 1: What happens here if you pass in a React component instead of a string?

timhowes-lab: If it's not a string then it will be picked up by the !isLongText check on line 12, and just return the content

packages/ui/cypress/component/reports/ExpandableCell.cy.tsx
timhowes-lab: Yeh fair point, shouldn't be mounting the parent row container structure

packages/ui/src/components/Reports/ExpandableCell.tsx
```diff
@@ -0,0 +1,28 @@
...
+  if (!isLongText) {
+    return <>{content}</>
+  }
+
+  return (
+    <div data-testid="expandable-cell">
+      <div data-testid="expandable-cell-text">
+        {isExpanded ? content : `${content.slice(0, longTextAbbreviationCharCount)}...`}
+      </div>
+      <ActionLink onClick={() => setIsExpanded(!isExpanded)} id={isExpanded ? "show-more-action" : "show-less-action"}>
```
smlbrd: Should these conditions be reversed?

Currently, if `isExpanded` is true, the id is set to `"show-more-action"` but the `ActionLink` text is set to `"show less"`

smlbrd (dismissed):

packages/ui/src/components/Reports/ExpandableCell.tsx
timhowes-lab: Yeh I'm not sure tbh, I took this logic from reallocate.tsx, I'm not sure if we want the id to reflect the current state of the cell i.e. when expanded, it's in the show-more-action state, or whether we want it to reflect what the next action would be if you click the button, I could be swayed either way...

smlbrd: I'm happy to defer to anyone else on this, my instinct is that because it's the `ActionLink` component, the id should reflect the action a user can take by clicking the action link, rather than the current state of the cell?

For example, if we're showing an expanded cell contents (e.g. `isExpanded` is `true`), the `ActionLink` text displayed is "show less", so the `ActionLink` id should be set to `"show-less-action"`?

timhowes-lab: Makes sense to me, I'll update it

packages/ui/cypress/component/reports/ExpandableCell.cy.tsx
```diff
@@ -0,0 +1,63 @@
+import ExpandableCell from "@/components/Reports/ExpandableCell"
+
+describe("<ReportTableRow />", () => {
```
richrace: This needs changing

richrace (approved):

PR #1985, opened 2026-05-18
User detail report changes

timhowes-lab:
-Updated inner/child tables of nested table to use an h4 header
-Updated mapToUserPerformanceDetailDtoDay to assign user's full name to fullName property if available, otherwise fallback to username
-Updated createReportCSV function to allow for nested tables, group headings will now be formatted 
-Separated out table structure generation from the table components and the csv generation
-Changed names of types for group/nested tables - a group report is now passed generic TTable, as it is a set of tables. A nested report is now passed generic TGroup and TTable, as it is a set of groups of tables


packages/ui/src/utils/tables/groupTable.ts
```diff
@@ -0,0 +1,32 @@
...
+    const dataList = isRecordArray(rawDataList) ? rawDataList : []
+    const cleanRows = dataList.filter(isRecord)
+
+    return {
+      formattedTableName: config.formatter ? formatGroupName(config, tableName) : tableName,
+      tableName: tableName,
+      rows: cleanRows as TRow[],
+      totals,
+      columns: config.columns
+    } as ReportTable<TRow>
```
richrace: I agree with the SonarCloud warning

packages/ui/src/utils/tables/nestedTable.ts
```diff
@@ -0,0 +1,58 @@
...
+        columns: mappedColumns
+      } as ReportTable<TRow>
+    })
+
+    return {
+      groupName,
+      formattedGroupName: config.formatter ? formatGroupName(config, groupName) : groupName,
+      totals: groupTotals,
+      tables
+    } as ReportTableGroup<TRow>
```
richrace: I agree with SonarCloud here as well

packages/ui/src/components/Reports/ReportTableBody.tsx
```diff
@@ -1,15 +1,15 @@
-import { BaseReportColumn } from "types/reports/Columns"
-import { ReportTableRow } from "components/Reports/ReportTableRow"
 import { StyledReportTableBody } from "components/Reports/ReportTableBody.styles"
+import { ReportTableRow } from "components/Reports/ReportTableRow"
+import { BaseReportColumn } from "types/reports/Columns"
 
 interface TableBodyProps<T> {
-  rows: T[]
+  rows?: T[]
```
richrace: Why is this now optional?

packages/ui/src/services/reports/csvGeneration/createReportCsv.ts
```diff
@@ -0,0 +1,44 @@
...
+  GroupedReportConfig,
+  NestedGroupedReportConfig,
+  ReportConfig
+} from "types/reports/Config"
+import { getFlatReportCsvChunks } from "./getFlatReportCsvChunks"
+import { getGroupReportCsvChunks } from "./getGroupReportCsvChunks"
+import { getNestedReportCsvChunks } from "./getNestedReportCsvChunks"
+
+export const createReportCsv = async (
+  parsedData: Record<string, unknown>[],
```
richrace: We don't need to cast to any `ReportConfig`. We just need to change the `parsedData`. It requires some refactoring.

richrace (changes requested):

packages/ui/src/services/reports/csvGeneration/createReportCsv.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      options
+    })) as unknown as typeof Blob
+  })
+
+  afterAll(() => {
+    global.Blob = originalBlob
+  })
+
+  it("should route to getFlatReportCsvChunks when structure is flat", async () => {
+    const config = { structure: "flat" } as any
```
richrace, line 39: Suggestion: do not use `any` type even in tests.

richrace (approved):

PR #1986, opened 2026-05-18
Use more explicit types for the Report Tables

richrace:


packages/ui/src/utils/reports/calculateTotalsForFlatStructure.test.ts
```diff
@@ -0,0 +1,60 @@
+import { calculateTotalsForFlatStructure } from "./calculateTotalsForFlatStructure" // Adjust the import path as needed
```
tomwatkins1994: Comment left in?

packages/ui/src/components/Reports/SimpleTable.tsx
```diff
@@ -1,21 +1,44 @@
...
+  config,
+  rows,
+  tableName,
+  nested
+}: SimpleTableProps<T>) => {
   if (config.structure !== "flat") {
     return null
   }
 
+  const totals = calculateTotalsForFlatStructure(rows, config.totalsConfig, config.calculateTotalsCallback)
```
tomwatkins1994: Should we memoise the totals here?

packages/ui/src/utils/reports/calculateTotalsForFlatStructure.test.ts
richrace: Whoops

packages/ui/src/components/Reports/SimpleTable.tsx
richrace: I always forget `useMemo` 🤦‍♂️

tomwatkins1994 (approved):

PR #1998, opened 2026-05-21
BICAWS7-4219 - Use Node.js 24

tomwatkins1994:
* Use new base images
* Update `.nvmrc`
* Correct Nginx path as new base images have changed this


timhowes-lab (dismissed):

.circleci/continue_config.yml
```diff
@@ -104,6 +104,8 @@ commands:
 
   build_audit_logging_images:
     steps:
+      - fetch_images:
+          images: nodejs-20-2023 nginx-nodejs-20-2023-supervisord
```
richrace, line 108: This needs updating?

richrace (approved):

.circleci/continue_config.yml
tomwatkins1994: The audit log API still uses the old images so I need to pull those in for that

PR #2007, opened 2026-05-26
Bicaws7 4260 UI collapsible groups and tables

timhowes-lab:


packages/ui/src/components/Reports/CollapsibleContainer.styles.tsx
```diff
@@ -0,0 +1,16 @@
+import styled from "styled-components"
+import { gdsTagBlue, tagBlue } from "utils/colours"
+
+interface HeaderButtonProps {
+  $clickable?: boolean
+  $headerType: "h3" | "h4"
+}
+
+const HeaderButton = styled.button<HeaderButtonProps>`
+  background-color: ${({ $headerType }) => ($headerType === "h3" ? tagBlue : gdsTagBlue)};
```
richrace, line 10: I like this. You can remove the `padding` and `background-color` from this: https://github.com/ministryofjustice/bichard7-next-core/blob/1c26c0a958fbf5eca80c17e04a29613a9b6c440f/packages/ui/src/components/Reports/GroupTable.styles.tsx#L4-L16

richrace (changes requested):

richrace (approved):

PR #2015, opened 2026-05-29
Bicaws7 4280 api the next court appearance is just the code

timhowes-lab:


packages/api/src/useCases/cases/reports/utils/formatOffenceData.ts
tomwatkins1994, line 1: Should this fall back to the code if for some reason the name is not found?

packages/ui/src/utils/reports/validateDateRange.test.ts
```diff
@@ -112,10 +112,10 @@ describe("validateDateRange", () => {
     })
 
     it("returns no errors for a mid-month start with dateTo exactly 1 month ahead", () => {
-      const from = subMonths(today, 1)
-      const to = addMonths(from, 1)
+      const midMonth = new Date(2026, 3, 15)
```
tomwatkins1994: Won't this test eventually fail as the date will be too far into the past?

timhowes-lab: Yep good point... have changed to be based around current date

tomwatkins1994 (approved):

PR #2024, opened 2026-06-05
Bicaws7 4197 create an endpoint to get a list of user lookups

timhowes-lab:


packages/common/types/UserGroup.ts
```diff
@@ -4,6 +4,7 @@ export enum UserGroup {
   AuditLoggingManager = "Audit Logging Manager",
   ExceptionHandler = "Exception Handler",
   GeneralHandler = "General Handler",
+  NewBichardUI = "New Bichard UI",
```
richrace, line 7: Why do we need to add this back in?

packages/api/src/useCases/users/fetchUserLookupList.ts
```diff
@@ -0,0 +1,38 @@
+import type { ApiUserLookupQuery } from "@moj-bichard7/common/types/ApiUserLookupQuery"
```
richrace: Where is the tests for this?

packages/api/src/useCases/users/fetchUserLookupList.ts
```diff
@@ -0,0 +1,38 @@
...
+): PromiseResult<UserLookupList> => {
+  if (!userAccess(user)[Permission.CanListUsers]) {
+    return new NotAllowedError()
+  }
+
+  const users = await fetchUserLookups(database, user, query?.usernameOrName)
+
+  if (isError(users)) {
+    logger.error(users)
+    throw users
```
richrace: Using `PromiseResult` we should not `throw` the error, we should `return` it.

packages/api/src/services/db/users/fetchUserLookups.ts
```diff
@@ -0,0 +1,60 @@
...
+    return { users: [] }
+  }
+
+  const forceClauses = user.visibleForces.map((f) => {
+    const trimmedForceCode = f.replace(/^0+(\d+)/, "$1")
+
+    const visibleForce = String.raw`\y0+${trimmedForceCode}\y`
+
+    return sql`u.visible_forces ~ ${visibleForce}`
+  })
```
richrace: We have functions to search for `visible_forces`:

https://github.com/ministryofjustice/bichard7-next-core/blob/8db159fe86168f83228579602ac56258de584d5d/packages/api/src/services/db/visibleForcesSql.ts

https://github.com/ministryofjustice/bichard7-next-core/blob/8db159fe86168f83228579602ac56258de584d5d/packages/api/src/services/db/mapUserRowToUser.ts#L16

https://github.com/ministryofjustice/bichard7-next-core/blob/8db159fe86168f83228579602ac56258de584d5d/packages/api/src/services/formatForceNumbers.ts

And 

https://github.com/ministryofjustice/bichard7-next-core/blob/f21c619addc64c6270638818ac05cc069efeedf1/packages/api/src/services/db/users/fetchUsers.ts#L19-L33

We should refactor this.

packages/common/types/UserGroup.ts
timhowes-lab: The /users endpoint is returning a schema error when I call it using the sample seed-data, I think because these users still contain New Bichard UI as a user group. I remember running into a similar issue back when we did some work to remove the New UI group. We ended up keeping some references to the New UI group rather than making a migration to remove it from the db entirely

packages/api/src/services/db/users/fetchUserLookups.ts
timhowes-lab: I've now separated this logic out into filterUsersByVisibleForces.ts

packages/common/contracts/ApiUserLookupQuery.ts
```diff
@@ -0,0 +1,7 @@
+import z from "zod"
```
richrace, line 1: We add this in the `common/contracts` folder rather than `types`

richrace (approved):

PR #2035, opened 2026-06-11
Update LEDS tests snapshots and tags

emadkaramad:


packages/e2e-test/utils/LedsApi/helpers/LedsTestApiHelper/retryRequest.ts
```diff
@@ -0,0 +1,27 @@
...
+import { delay } from "../../../puppeteer-utils"
+
+const MAX_ATTEMPTS = 5
+const WAIT_BETWEEN_ATTEMPTS_IN_SECONDS = 10
+
+const retryRequest = async <T extends AxiosResponse>(request: () => Promise<T>): Promise<T | AxiosError> => {
+  let attempts = 0
+  while (true) {
+    const response = await request().catch((error: AxiosError) => error)
+    if (response.status === HttpStatusCode.BadGateway) {
```
tausifpatel, line 12: Don't we need to cover other server errors? Such as 503 and 504 etc.

emadkaramad: We can, but it takes about a minute for the Gateway Timeout error to come back, and retrying it would use up quite a bit of pipeline time. Let's see how it goes, and then we can adjust it based on the result.

emadkaramad: The cucumber test itself is set to retry 5 times already, so for 503 (Service Unavailable) I think the 5 times retry should be fine. We haven't received that error recently anyway.

tausifpatel (approved):

PR #2037, opened 2026-06-12
feat: add Zod validation to fetchAudit

Dtzz81:


packages/api/src/services/db/audit/fetchAudit.ts
tomwatkins1994: We should remove this type here as we will now be relying on Zod to get the type for us.

packages/api/src/services/db/audit/fetchAudit.ts
```diff
@@ -41,5 +42,10 @@ export const fetchAudit = async (
     return null
   }
 
-  return convertAuditToDto(results[0])
+  const parsedResults = z.array(AuditSchema).safeParse(results)
```
tomwatkins1994: To keep things simple I think we should just do

```js
const parsedResults = AuditSchema.safeParse(results[0])
```

tomwatkins1994 (changes requested):

tomwatkins1994 (approved):

PR #2038, opened 2026-06-12
Bicaws7 4198 create an endpoint to lock a case with a user

timhowes-lab:
- Added getUser endpoint and fetchUserById function
- Added allocate endpoint to lock triggers/exceptions for a given case and user
- Updated UI to link up with new lock endpoint


packages/api/src/routes/v1/cases/[caseId]/allocate.e2e.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    const [encodedJwt] = await createUserAndJwtToken(helper.postgres, [UserGroup.Audit], {
+      visibleForces: ["01"]
+    })
+    const caseObj = await createCase(helper.postgres, { errorLockedById: null })
+
+    const targetUser = await createUser(helper.postgres)
+
+    const url = `${helper.address}${getRouteUrl(caseObj.errorId)}?allocatedToUserId=${targetUser.id}&caseType=exceptions`
+    const response = await fetch(url, defaultRequest(encodedJwt))
+    console.log(response)
```
Comment (richrace) [inline packages/api/src/routes/v1/cases/[caseId]/allocate.e2e.test.ts]: Remove console

packages/api/src/routes/v1/cases/[caseId]/allocate.e2e.test.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    expect(response.status).toBe(FORBIDDEN)
+  })
+
+  it("returns 500 INTERNAL SERVER ERROR if the target user being allocated to does not exist", async () => {
```
Comment (richrace) [inline packages/api/src/routes/v1/cases/[caseId]/allocate.e2e.test.ts]: Should this be `404 Not found error`? We have this here: https://github.com/ministryofjustice/bichard7-next-core/blob/main/packages/api/src/types/errors/NotFoundError.ts

packages/api/src/routes/v1/cases/[caseId]/allocate.e2e.test.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    const caseObj = await createCase(helper.postgres, { errorLockedById: null })
+    const url = `${helper.address}${getRouteUrl(caseObj.errorId)}?allocatedToUserId=99999&caseType=exceptions`
+
+    const response = await fetch(url, defaultRequest(encodedJwt))
+
+    expect(response.status).toBe(INTERNAL_SERVER_ERROR)
+  })
+
+  it("returns 500 INTERNAL SERVER ERROR if the target user exists but falls outside the actor's visible forces", async () => {
```
Comment (richrace) [inline packages/api/src/routes/v1/cases/[caseId]/allocate.e2e.test.ts]: This should `403 Forbidden`

packages/api/src/routes/v1/cases/[caseId]/allocate.e2e.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    expect(response.status).toBe(OK)
+
+    const updatedCase = await helper.postgres.writable
+      .connection`SELECT * FROM br7own.error_list WHERE error_id = ${caseObj.errorId}`
+
+    expect(updatedCase[0].error_locked_by_id).toBe(targetUser.username)
+    expect(updatedCase[0].trigger_locked_by_id).toBeNull()
+  })
+
+  it("returns 500 INTERNAL SERVER ERROR if downstream lockAndAuditLog execution fails due to a missing case entry", async () => {
```
Comment (richrace) [inline packages/api/src/routes/v1/cases/[caseId]/allocate.e2e.test.ts]: What's the difference between this test and `returns 500 INTERNAL SERVER ERROR if the target user being allocated to does not exist`?

```typescript
const url = `${helper.address}${getRouteUrl(caseObj.errorId)}?allocatedToUserId=99999&caseType=exceptions`
``` 

Neither has a case been found.

packages/api/src/services/db/users/fetchUserById.integration.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  afterAll(async () => {
+    await testDatabaseGateway.close()
+  })
+
+  it("should return user if id is valid and user belongs to a visible force in the current user's visible forces", async () => {
+    const currentUser = await createUser(testDatabaseGateway, { visibleForces: ["01"] })
+    const requestedUser = await createUser(testDatabaseGateway, { visibleForces: ["01"] })
+
+    const result = (await fetchUserById(testDatabaseGateway.readonly, currentUser, requestedUser.id)) as User
```
richrace: Why do we need to cast here?

packages/api/src/services/db/users/fetchUserById.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  if (user.visibleForces.length === 0) {
+    return new Error(`User with ID "${id}" has no visible forces`)
+  }
+
+  const visibleForcesFilter = filterUsersByVisibleForces(database, user.visibleForces)
+
+  const visibleForcesWhere = sql`(${visibleForcesFilter})`
+
+  const userResult = await database.connection<UserRow[]>`
```
richrace: Do we need all of the user's information?

packages/api/src/routes/v1/cases/[caseId]/allocate.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    return reply.code(INTERNAL_SERVER_ERROR).send()
+  }
+
+  return reply.code(OK).send(result)
+}
+
+const route = async (fastify: FastifyInstance) => {
+  useZod(fastify).put(V1.CasesAllocate, { schema }, async (req, reply) => {
```
Comment (richrace) [inline packages/api/src/routes/v1/cases/[caseId]/allocate.ts:69]: We are using `PUT`, but we're updating the data via the query.

We should be using the body.

richrace (changes requested):

packages/api/src/routes/v1/cases/[caseId]/allocate.e2e.test.ts
Comment (timhowes-lab) [reply packages/api/src/routes/v1/cases/[caseId]/allocate.e2e.test.ts]: This test fails inside the lockAndAuditLog function due to a missing case, whereas the other fails before that function is called due to a missing user

packages/api/src/routes/v1/cases/[caseId]/allocate.ts
```diff
@@ -0,0 +1,82 @@
...
+  caseId: number
+  database: DatabaseGateway
+  logger: FastifyBaseLogger
+  reply: FastifyReply
+  user: User
+}
+
+const schema = {
+  ...auth,
+  body: AllocationQuerySchema,
```
Comment (richrace) [inline packages/api/src/routes/v1/cases/[caseId]/allocate.ts]: This name should not have `Query` suffix.

richrace (approved):

PR #2048, opened 2026-06-17
BICAWS7-4189-api-add-parsing-to-all-database-calls-rest

Dtzz81:
Add Zod parsing to all database calls in the API to verify the data returned (and also coerce values like dates). Some databases need more work; e.g., implement ZOD first. Hence, 2 new tickets were added for those.


Dtzz81:
old branch, not in use

PR #2056, opened 2026-06-18
Bicaws7 4248 show an error state when a user doesnt enter a valid string in the new force owner field in the case reallocation journey

timhowes-lab:


richrace (changes requested):
I can still submit the form without entering anything in, and I don't get an error.

richrace (approved):

PR #2060, opened 2026-06-19
BICAWS7-4271 - Redirect for any non-https request

katerd:
- add missing host `www.gsi.exchange.gov.uk` to redirect logic
- update redirect logic to use `globalThis` (assuming a window) instead.
- tests run under http://localhost so host check cannot be immediately be removed

### DIFF
--- a/packages/user-service/src/pages/login/index.tsx
+++ b/packages/user-service/src/pages/login/index.tsx
@@ -66,6 +66,26 @@ const getNotYourEmailLink = (query: ParsedUrlQuery): string => {
   })
 }
 
+const shouldRedirectToHttps = (location: Location, httpsRedirectCookie: boolean): boolean => {
+  if (!httpsRedirectCookie) {
+    return false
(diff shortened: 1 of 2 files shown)


packages/user-service/src/pages/login/index.tsx
```diff
@@ -65,6 +66,26 @@ const getNotYourEmailLink = (query: ParsedUrlQuery): string => {
...
+  }
+
+  const isSecure = location.protocol.includes("https")
+  if (isSecure) {
+    return false
+  }
+
+  const hosts = [
+    "bichard7.service.justice.gov.uk",
+    "proxy.bichard7.justice.gov.uk",
```
tomwatkins1994: This should be `proxy.bichard7.service.justice.gov.uk`

tomwatkins1994 (changes requested):

tomwatkins1994 (approved):

PR #2068, opened 2026-06-23
BICAWS7-4189-api-add-parsing-to-all-database-calls-audit-and-cases

Dtzz81:
Add Zod parsing to all database calls in the API to verify the data returned (and also coerce values like dates). Some databases need more work; e.g., implement ZOD first. Hence, 2 new tickets were added for those.


packages/api/src/services/db/cases/fetchCaseAges.e2e.test.ts
```diff
@@ -82,12 +82,12 @@ describe("fetchCasesAndFilter fetchCaseAges e2e", () => {
     const caseMetadata = (await fetchCasesAndFilter(helper.postgres.readonly, defaultQuery, user)) as CaseIndexMetadata
 
     expect(caseMetadata.cases).toHaveLength(14)
-    expect(caseMetadata.caseAges[CaseAge.Today]).toBe("4")
```
timhowes-lab, line 85: Wondering if we should cast this to an int in the count query in fetchCaseAges.ts using ::int

timhowes-lab (approved):

PR #2069, opened 2026-06-23
BICAWS7-4189-api-add-parsing-to-all-database-calls-fetch-by-user-Id

Dtzz81:
Add Zod parsing to  database calls in the API to verify the data being returned


packages/common/types/User.ts
```diff
@@ -17,6 +17,15 @@ export const UserRowSchema = z.object({
   visible_forces: z.string().nullable()
 })
 
+export const UserMinimalRowSchema = z.object({
```
tomwatkins1994: We could use the `z.pick({ ... })` method here to reduce duplication.

https://zod.dev/api?id=pick#pick

tomwatkins1994 (changes requested):

packages/common/types/User.ts
tomwatkins1994: We should use `z.infer` here to get the type from the `UserMinimalRowSchema`

tomwatkins1994 (changes requested):

packages/common/types/User.ts
```diff
@@ -17,6 +17,24 @@ export const UserRowSchema = z.object({
...
+export const UserMinimalRowSchema = UserRowSchema.pick({
+  deleted_at: true,
+  groups: true,
+  id: true,
+  username: true,
+  visible_courts: true,
+  visible_forces: true
+})
+
+// export const UserMinimalRowSchema = z.object({
```
bjpirt: We can get rid of the old commented code now

bjpirt (changes requested):

bjpirt (approved):

tomwatkins1994 (approved):

PR #2072, opened 2026-06-24
BICAWS7-4360-fetch-userby-username

Dtzz81:
Implemented parsing in UserRowSchema


packages/api/src/services/db/users/fetchUserByUsername.ts
```diff
@@ -37,8 +42,13 @@ export default async (database: DatabaseConnection, username: string): PromiseRe
   }
 
   if (!userResult || userResult.length === 0) {
-    return Error(`User "${username}" does not exist`)
+    return new NotFoundError()
```
timhowes-lab: My bad- this does need to stay as the original error type after all, because the verifyJwt integration test relies on it

katerd (approved):

PR #2075, opened 2026-06-25
Bicaws7 3538 add resolved by user filter to the filter panel for reports for resolved exceptions and triggers

timhowes-lab:
- Updated exceptions report endpoint to take 'resolved by' users as an optional query param, results are filtered using the users, or all results returned if no users specified.
- Moved 'resolved by' checkboxes logic out of audit search and into separate component that can be used in audit search and the reports page.
- Updated exceptions report UI page to include resolved by checkboxes.
- Reports page component tests were split across several files, have aggregated into one ReportsSelectionFilter.cy.tsx file. Previously it wasn't clear what component was being tested based off test file name


packages/ui/src/pages/report-selection/index.tsx
```diff
@@ -17,30 +21,49 @@ import AuthenticationServerSidePropsContext from "../../types/AuthenticationServ
...
 
 interface Props {
   user: DisplayFullUser
   canUseTriggerAndExceptionQualityAuditing: boolean
+  resolvedBy: AuditResolvedBy[]
 }
 
 function ReportSelectionPage(props: Readonly<Props>) {
-  const { canUseTriggerAndExceptionQualityAuditing, user } = props
+  const { canUseTriggerAndExceptionQualityAuditing, user, resolvedBy } = props
```
tomwatkins1994, line 66: I might be missing something here but why do we need to get the feature flag for the user to be able to do auditing on the reports page?

I can see we're drilling the value to other components but I can't work out why it is needed there either.

timhowes-lab: It's specifically for the 'Resolved by' filter on the Resolved Exceptions and Triggers report. If the user has trigger and exception quality auditing permission, Resolved by should be shown

image

richrace: This is also needed for the top nav, as it is a page.

richrace (approved):

PR #2078, opened 2026-06-30
Bicaws7 4368 add parsing to fetch user lookups

Dtzz81:
Updated convertUserToDto as this is connected to both fetchUserLookups and fetchUserLookupList (parsing will be implemented in that file in next PR).


packages/api/src/services/db/users/fetchUserLookups.ts
```diff
@@ -1,13 +1,15 @@
 import type { PromiseResult } from "@moj-bichard7/common/types/Result"
-import type { User, UserRow } from "@moj-bichard7/common/types/User"
+import type { UserLookupRow } from "@moj-bichard7/common/types/User"
+
+import { type User, UserLookupRowSchema } from "@moj-bichard7/common/types/User"
```
katerd:
```suggestion

import { type User, type UserLookupRow, UserLookupRowSchema } from "@moj-bichard7/common/types/User"
```

Would this work?

Dtzz81: lint complains I tried in VS code :( I tried it here directly, but I'm not sure.

katerd (approved):

PR #2090, opened 2026-07-03
BICAWS7-4380-get-potential-cases-to-audit

Dtzz81:
Implemented parsing


packages/common/types/Case.ts
```diff
@@ -97,6 +97,14 @@ export const CaseSchema = z.object({
...
 })
 
+export const casesToAuditSchema = CaseSchema.pick({
+  errorId: true,
+  errorQualityChecked: true,
+  errorResolvedBy: true,
+  triggerQualityChecked: true,
+  triggerResolvedBy: true
+})
+
```
katerd: This should probably live within `types/Audit.ts` - and be named `CaseForAuditSchema`, as its purpose is for the audit functionality.

packages/api/src/useCases/dto/convertCasestoAuditDto.ts
```diff
@@ -0,0 +1,8 @@
+export const convertCasesToAudit = (rows: Record<string, unknown>[]) =>
+  rows.map((row) => ({
+    errorId: row.error_id,
+    errorQualityChecked: row.error_quality_checked,
+    errorResolvedBy: row.error_resolved_by,
+    triggerQualityChecked: row.trigger_quality_checked,
+    triggerResolvedBy: row.trigger_resolved_by
+  }))
```
katerd: If this function is only to be used by a single location, I'd probably move it into the getPotentialCasesToAudit file (maybe even inline this map) - especially as we're dealing with `unknown` - reducing the spread of loosely defined types reduces the chance of mistakes

packages/common/types/Case.ts
katerd: I used `CaseRowForBailsReportSchema` and `CaseRowForWarrantsReportSchema` for examples where `.pick` is used in different pieces of functionality

packages/api/src/services/db/audit/getPotentialCasesToAudit.ts
```diff
@@ -82,21 +84,41 @@ export async function getPotentialCasesToAudit(
     return new Error("Failed to get cases to audit")
   }
 
+  const auditMappingSchema = z.preprocess((val) => {
+    if (!val || typeof val !== "object") {
+      return val
+    }
+
+    const row = val as Record<string, unknown>
```
bjpirt: It feels like we're missing the benefit of zod here with this manual cast. Shouldn't we either define a schema for the database rows with the columns matching the name (snake_case) or alias the columns as we make the SQL query (e.g. `SELECT error_id as errorId`). But turning this into an unknown feels wrong here

packages/common/types/Audit.ts
```diff
@@ -13,6 +15,14 @@ export const AuditSchema = z.object({
   volume_of_cases: z.number()
 })
 
+export const casesToAuditSchema = CaseSchema.pick({
```
tomwatkins1994: Could we capitalise this pls `casesToAuditSchema` -> `CasesToAuditSchema`

tomwatkins1994: Could this also be called `CasesToAuditRowSchema` and `.pick` from `CaseRowSchema` instead of `CaseSchema`? We can then use the `snake_case` field names and therefore not need to use SQL field aliases like `error_id AS "errorId"`.

This will make it more consistent with how we have done SQL calls elsewhere.

packages/api/src/services/db/audit/getPotentialCasesToAudit.ts
```diff
@@ -71,32 +73,37 @@ export async function getPotentialCasesToAudit(
 
   const results = await sql<
     {
```
tomwatkins1994: We can remove this type because it will be inferred from the Zod schema

packages/api/src/services/db/audit/getPotentialCasesToAudit.ts
```diff
@@ -22,11 +24,11 @@ export async function getPotentialCasesToAudit(
 
   const baseQuery = sql`
     SELECT 
-      error_id,
-      trigger_resolved_by,
-      trigger_quality_checked,
-      error_resolved_by,
-      error_quality_checked
+      error_id AS "errorId",
```
tomwatkins1994: Remove aliases as described in other comments

tomwatkins1994 (changes requested):

tomwatkins1994 (approved):

PR #2123, opened 2026-07-22
BICAWS7-4402 Use next hearing location instead of code

CaitBarnard:
Currently the Next Hearing Location field shows just the Organisation Unit code (e.g B37HI00).
We've updated it so that it now shows code and name (e.g  B37HI00 - Magistrates' Courts Suffolk Ipswich)
This is also updated in the typeahead component

In order to get the full name of a court, given a code, NextHearingLocationField.tsx needs access to the full list of organisation units. We added OrganisationUnitsContext, this stores all organisation units, fetched in the court case server side props call. This is then used in NextHearingLocationField.tsx


packages/ui/src/components/Typeaheads/OrganisationUnitTypeahead.tsx
```diff
@@ -72,7 +73,12 @@ const OrganisationUnitTypeahead: React.FC<Props> = ({
       }
     },
     initialInputValue: value,
-    itemToString: (item) => item?.fullOrganisationCode ?? ""
+    itemToString: (item) => {
+      if (item) {
+        return `${item.fullOrganisationCode} - ${item.fullOrganisationName}`
+      }
+      return ""
+    }
```
richrace: Can condense this:

```typescript
itemToString: (item) => item ? `${item.fullOrganisationCode} - ${item.fullOrganisationName}` : ""
```

packages/ui/src/components/Typeaheads/OrganisationUnitTypeahead.tsx
```diff
@@ -62,7 +63,7 @@ const OrganisationUnitTypeahead: React.FC<Props> = ({
       amend("nextSourceOrganisation")({
         resultIndex: resultIndex,
         offenceIndex: offenceIndex,
-        value: inputValue
+        value: inputValue.split(" ")[0]
```
richrace: Suggestion/opinion: Split the value above this

packages/ui/src/features/CourtCaseDetails/Tabs/Panels/EditableFields/NextHearingLocationField.tsx
```diff
@@ -29,27 +29,71 @@ export const NextHearingLocationField = ({
   const { amendments } = useCourtCase()
   const amendedNextHearingLocation = getNextHearingLocationValue(amendments, offenceIndex, resultIndex) ?? ""
   const [isNhlSaved, setIsNhlSaved] = useState<boolean>(false)
-  const [organisations, setOrganisations] = useState<OrganisationUnitApiResponse>([])
+  const [organisations, setOrganisations] = useState<OrganisationUnitApiResponse | undefined>(undefined)
```
richrace: Question: Why is it now `undefined`?

CaitBarnard: Organisations is initially empty until the useEffect runs. Once it's no longer undefined, the Typeahead is able to get the correct code/name.

packages/common/utils/searchCourtOrganisationUnits.test.ts
```diff
@@ -1,11 +1,12 @@
 import searchCourtOrganisationUnits, {
   getFullOrganisationName,
-  getOrganisationCodeAndName
+  getOrganisationCodeAndName,
+  sortedCourtOrganisationUnits
 } from "./searchCourtOrganisationUnits"
 
 describe("searchCourtOrganisationUnits", () => {
   it("Should return an empty array when search keyword is an empty string", () => {
```
CaitBarnard:
```suggestion
  it("Should return an array with all organisation units when search keyword is an empty string", () => {
```

packages/ui/src/features/CourtCaseDetails/Tabs/Panels/EditableFields/NextHearingLocationField.tsx
```diff
@@ -27,32 +27,43 @@ export const NextHearingLocationField = ({
...
+  const { organisationUnits } = useOrganisationUnits()
   const amendedNextHearingLocation = getNextHearingLocationValue(amendments, offenceIndex, resultIndex) ?? ""
   const [isNhlSaved, setIsNhlSaved] = useState<boolean>(false)
-  const [organisations, setOrganisations] = useState<OrganisationUnitApiResponse>([])
   const [isNhlChanged, setIsNhlChanged] = useState<boolean>(false)
 
-  const isValidNhl = isValidNextHearingLocation(amendedNextHearingLocation, organisations)
+  const originalCode = result.NextResultSourceOrganisation?.OrganisationUnitCode
 
+  const isValidNhl = isValidNextHearingLocation(amendedNextHearingLocation, organisationUnits)
```
emadkaramad, line 37: nitpick: We usually use the full names for variables (there are some exceptions like Aho, Pnc, etc):
`isValidNextHearingLocation` instead of `isValidNhl`

CaitBarnard: I think it was originally done like that as `isValidNextHearingLocation` clashes with the validator of the same name. So we can always change `isValidNhl` to `isValidNextHearingLocation` and use the validator as `isValidNextHearingLocationValidator` instead?

packages/ui/cypress/component/Typeaheads/OrganisationUnitTypeahead.cy.tsx
```diff
@@ -87,22 +87,6 @@ describe("OrganisationUnitTypeahead Component", () => {
     cy.get("input#next-hearing-location").should("be.visible").and("have.value", "B01EF00")
   })
 
-  it("triggers network request and sets parent organization lists on load/input", () => {
```
richrace, line 90: Question: Why has this test been deleted?

packages/ui/src/pages/court-cases/[courtCaseId]/index.tsx
```diff
@@ -267,6 +273,11 @@ export const getServerSideProps = withMultipleServerSideProps(
       ? (apiCase as DisplayFullCourtCase)
       : courtCaseToDisplayFullCourtCaseDto(courtCase as CourtCase, currentUser)
 
+    const organisationUnits: OrganisationUnitApiResponse = searchCourtOrganisationUnits("").map((ou) => ({
+      fullOrganisationCode: getFullOrganisationCode(ou),
+      fullOrganisationName: getFullOrganisationName(ou)
+    }))
```
Comment (richrace) [inline packages/ui/src/pages/court-cases/[courtCaseId]/index.tsx:279]: Question: I don't understand this.

packages/ui/src/features/CourtCaseDetails/Tabs/Panels/EditableFields/NextHearingLocationField.tsx
```diff
@@ -27,32 +27,43 @@ export const NextHearingLocationField = ({
...
   const isEditable = isCaseEditable && hasNextHearingLocationException(exceptions)
 
+  const getDisplayValue = (code?: string | null) => {
+    if (!code) {
+      return ""
+    }
+    const matchingOrg = organisationUnits.find((org) => org.fullOrganisationCode === code)
+    return matchingOrg ? `${code} - ${matchingOrg.fullOrganisationName}` : code
+  }
```
richrace, line 46: Doesn't display that it is not an unknown code

packages/ui/src/pages/court-cases/[courtCaseId]/index.tsx
Comment (timhowes-lab) [reply packages/ui/src/pages/court-cases/[courtCaseId]/index.tsx:279]: So initially I changed NextHearingLocationField.tsx to make an API call to organisation-units.ts on render, to pull in all org units. I had a return statement that prevented the component being rendered until the org units had been loaded.

However, this was causing some cucumber tests to fail, because they were unable to find the input component, as the org units list had not been pulled in yet.

So I've added a new context to store the org units, which is populated in the server side props call in this index.tsx file, and then the org units are accessed via a hook in NextHearingLocationField.tsx.

Comment (richrace) [reply packages/ui/src/pages/court-cases/[courtCaseId]/index.tsx:279]: Got it 👍 

We should add `DisplayName` to the context.

Screenshot 2026-08-03 at 15 21 42

packages/ui/cypress/component/Typeaheads/OrganisationUnitTypeahead.cy.tsx
timhowes-lab: Re-added but removed the setOrganisations function call checks, as this component now doesn't set the org units list in the parent

richrace (approved):

PR #2131, opened 2026-07-27
Bicaws7 4395 UI we should start the trace id before the api call

timhowes-lab:
ApiClient and ReportsApiClient now take traceId as a parameter, rather than generating inside their request methods.

### DIFF
--- a/packages/ui/src/services/api/ApiClient.ts
+++ b/packages/ui/src/services/api/ApiClient.ts
@@ -1,8 +1,8 @@
 import { API_LOCATION } from "config"
+import { randomUUID } from "node:crypto"
 import { ApiError } from "types/ApiError"
 import type PromiseResult from "types/PromiseResult"
 import { Agent, fetch } from "undici"
-import { randomUUID } from "node:crypto"
 import apiLogger from "./apiLogger"
 
 export enum HttpMethod {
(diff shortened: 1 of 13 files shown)


packages/ui/src/services/api/ReportsApiClient.ts
```diff
@@ -18,7 +20,7 @@ export default class ReportsApiClient {
   }
 
   async *fetchReport<T>(url: string): AsyncIterable<T | Error> {
-    const traceId = randomUUID()
+    const traceId = this.traceId
     const logger = apiLogger(traceId, url)
```
richrace: This can be `const logger = apiLogger(this.traceId, url)`

packages/ui/src/services/api/ApiClient.ts
```diff
@@ -42,7 +44,7 @@ class ApiClient {
     bodyContent: string | Record<string, unknown> = {},
     additionalHeaders: Record<string, unknown> = {}
   ): PromiseResult<T> {
-    const traceId = randomUUID()
+    const traceId = this.traceId
     const logger = apiLogger(traceId, route)
```
richrace: Same again

richrace (changes requested):

packages/ui/src/services/api/ReportsApiClient.ts
timhowes-lab: Yeh I kept the local traceId var because there's a few other places in the function where the traceId is referenced, and I thought it would be cleaner than swapping for this.traceId for each occurrence. I'm fine with making the switch though

richrace (approved):

PR #2139, opened 2026-07-31
BICAWS7-4310-Test the container in-CI with Goss

Dtzz81:


packages/s3-web-proxy/s3-server/package.json
```diff
@@ -10,5 +10,11 @@
     "lint": "exit 0",
     "test": "jest --passWithNoTests",
     "test:goss": "bash ../build.sh"
+  },
+  "dependencies": {
```
tomwatkins1994, line 38: Where have these come from?

packages/s3-web-proxy/Makefile
tomwatkins1994, line 1: Could we add a fallback for the `DOCKER_TAG` so this `test-goss` runs locally?

`DOCKER_TAG ?= latest`

packages/s3-web-proxy/Makefile
```diff
@@ -1,5 +1,14 @@
...
 	bash build.sh
+
+# Install goss and dgoss using the script
+install-goss:
+	sudo bash ../../scripts/install-goss.sh
+
+# Run goss container tests
+test-goss:
+	GOSS_FILE=packages/s3-web-proxy/goss.yaml GOSS_SLEEP=5 \
+	dgoss run -e S3_BUCKET_NAME="test-bucket" "s3-web-proxy:$(DOCKER_TAG)"
```
tomwatkins1994: This seems to be missing some of the environment variables from `build.sh` like `AWS_ACCESS_KEY_ID` etc.

packages/s3-web-proxy/Makefile
```diff
@@ -1,5 +1,14 @@
...
 build:
 	bash build.sh
+
+# Install goss and dgoss using the script
+install-goss:
+	sudo bash ../../scripts/install-goss.sh
+
+# Run goss container tests
+test-goss:
+	GOSS_FILE=packages/s3-web-proxy/goss.yaml GOSS_SLEEP=5 \
```
tomwatkins1994: Could we instead use `GOSS_FILES_PATH=packages/s3-web-proxy` so it find the `goss_wait.yaml` too?

packages/s3-web-proxy/Makefile
```diff
@@ -1,5 +1,14 @@
...
 
 build:
 	bash build.sh
+
+# Install goss and dgoss using the script
+install-goss:
+	sudo bash ../../scripts/install-goss.sh
+
+# Run goss container tests
+test-goss:
```
tomwatkins1994: Could we just rename this from `test-goss` -> `goss` to match other Makefiles in the repo see [packages/ui/Makefile](https://github.com/ministryofjustice/bichard7-next-core/blob/1b745617d8548834e28bd420288bab9894af1e41/packages/ui/Makefile)

.circleci/continue_config.yml
```diff
@@ -503,6 +509,16 @@ jobs:
...
+      - run:
+          name: Run Goss Tests
+          command: |
+            #!/bin/bash -eo pipefail
+            source $BASH_ENV
+
+            export GOSS_FILE=packages/s3-web-proxy/goss.yaml
+            export GOSS_SLEEP=5
+
+            dgoss run -e S3_BUCKET_NAME="test-bucket" "s3-web-proxy:${DOCKER_TAG}"
```
tomwatkins1994: We have the command in the Makefile but we don't seem to be using it here?

We will need to `cd` into the right directory and run the the new `make goss` command

.circleci/continue_config.yml
```diff
@@ -488,6 +488,12 @@ jobs:
           steps:
             - run: circleci-agent step halt
       - checkout
+      - run:
+          name: Install Goss
+          command: |
+            sudo bash ./scripts/install-goss.sh
+            echo 'export GOSS_PATH="/usr/local/bin/goss"' >> $BASH_ENV
```
tomwatkins1994: This line is already at the end of the `install-goss.sh` file so I don't think we need it here

.circleci/continue_config.yml
```diff
@@ -503,6 +508,14 @@ jobs:
           command: |
             mkdir -p images
             docker image save "s3-web-proxy:${DOCKER_TAG}" | pigz > "images/s3-web-proxy.tar.gz"
+      - run:
+          name: Run Goss Tests
+          command: |
+            #!/bin/bash -eo pipefail
```
tomwatkins1994: I think we can get rid of these lines too

```
#!/bin/bash -eo pipefail
source $BASH_ENV
```

